PR #3, opened 2013-10-15
adding unitedstates to list of civic hacker orgs

konklone:


benbalter:
Great catch! Thanks. :us: :us:

PR #6, opened 2013-10-15
Reworded title and body to emphasize the multilateral nature of the project

konklone:
Giving credit to GovTrack and the New York Times where it's due, along with a couple other copy tweaks.


benbalter:
Mea culpa again for not properly emphasizing the distinction. 

@joshdata, @dwillis does that more accurately reflect the project contributors?

dwillis:
That is much better, thank you.

PR #7, opened 2013-10-15
Update link to repo in FAQ

sshirokov:
The organization name was missing from the URL, leading to a 404.


gjtorikian:
@benbalter How'd HTML-Proofer not catch this?

PR #11, opened 2013-10-15
Addition of LLNL Chaos account.

bringhurst:
A group of talented folks who work at llnl.gov.


jlord:
Thanks @fintler! So awesome to learn about these organizations!

PR #14, opened 2013-10-15
Addition of hpc account.

bringhurst:
The hpc account is used (almost exclusively) by US national labs.


jlord:
Super excited about this one! Thanks @fintler!

PR #16, opened 2013-10-15
Use "embed" instead of "imbed"

troyswanson:
Not sure which is preferable here, but I believe "embed" is the main word and "imbed" is the variant.


afeld:
@troyswanson Thanks!

troyswanson:
:heart:

PR #19, opened 2013-10-15
Update _config.yml to add City of Greater Sudbury

cschiewek:
Added the City of Greater Sudbury's github account to _config.yml.  The City of Greater Sudbury is a municipality located in Northern Ontario, Canada. 

http://www.greatersudbury.ca


jlord:
Thanks @cschiewek!

PR #21, opened 2013-10-15
Added cleveland metroparks

smathermather:
Adding Cleveland Metroparks as a gov agency in the list of USA agencies.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -5,7 +5,6 @@ organizations:
     - adlnet
     - cfpb
     - chaos
-    - cleveland-metroparks
     - cmsgov
     - commercegov
     - department-of-veterans-affairs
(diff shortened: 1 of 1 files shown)


jlord:
Thanks @smathermather! Do you think it should go under US City though?

smathermather:
Hmm, good question.  We're a "special district" or subdivision of the state, but we aren't state government either... .  Let me take another look.

smathermather:
See if this makes more sense...

jlord:
Thanks @smathermather! I'm good with that, hopefully there will be others in that group. As these pull requests come in we'll get a better idea of the scope and the subdivisions that make sense for organizing everyone. I'm super excited how many have come in!

smathermather:
Thanks @jlord.  Re-arrange as you see fit.  "Special District" is very special and uncommon, even within the US, so maybe "Local" would be a better/more generic fit... .  Or I can change and redo the pull request.

PR #22, opened 2013-10-15
Added Utah's GIS shop to US States

steveoh:
added the Utah Automated Geographic Reference center. http://gis.utah.gov


jlord:
Yeah GIS! Thanks @steveoh!

PR #23, opened 2013-10-15
Renaming file/URL for modern approach post

konklone:
One thing I didn't notice in my last PR, was the URL for the story is just "sunlight". This renames the filename/slug to use the title of the post. I updated the Rakefile as well, and couldn't find any other hardcoded places the old filename was mentioned.


konklone:
The previous PR I mentioned is #6, for reference. Its point was to make it clear the project in question wasn't just a Sunlight effort. Switching the URL from "sunlight" to something else is helpful for that.

afeld:
I saw a lot of tweets today linking to this article specifically, so we don't want to break those.  If you can get both to work using https://github.com/edsl/jekyll-rename-posts or something, feel free to re-open.  Thanks!

konklone:
Yeah, I should have handled that in my PR, good call. I think a 301 redirect would be better, and [search engines handle meta refreshes with 0 delay as 301's](http://sebastians-pamphlets.com/google-and-yahoo-treat-undelayed-meta-refresh-as-301-redirect/), so I'll add a tiny page with a meta tag to the PR and re-open.

benbalter:
Read the threads out of order. @afeld, @konklone, you're absolutely right. Didn't want to add a new post, but also didn't want to keep the bad url out there and make the damage worse.

Also, 0 delays are 301s? 

The more you know

PR #29, opened 2013-10-15
Update _config.yml

sebhoss:
Added german @bundestag

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -67,6 +67,9 @@ organizations:
   France:
     - etalab
 
+  Germany:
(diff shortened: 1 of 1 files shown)


jlord:
Oh yeah, I love this - it's inspired a bunch of stuff here in the states. It's technically not from the government though, correct? So we should put it in the Civic Hackers section then?

sebhoss:
Right, it's not run by the government. Moved to civic hackers.

jlord:
Awesome! Thanks @sebhoss!

PR #30, opened 2013-10-15
Adds erasme

leucos:
Added erasme (local gov agency) in France section

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -65,6 +65,7 @@ organizations:
     - City-of-Helsinki
 
   France:
+    - erasme
     - etalab
 
   Norway:


jlord:
Great! @leucos is this an organization of the government or of local civic hackers/volunteers?

leucos:
Hi @jlord ! 
ERASME  is a unit working on innovation regarding education, culture, people with disabilities, elders. we're part of a local govement called  "Département" which is a kind of "State" here.
Our home : http://www.erasme.org
Our mothership : http://www.rhone.fr

jlord:
Thanks for submitting @leucos!

PR #31, opened 2013-10-15
Change references of gov.github.com to government.github.com

AnSavvides:
The title says it all; using gov.github.com leads to a 404 (the same applies for anything repository related that uses gov.github.com).


jlord:
Eek! Thanks so much for catching these @AnSavvides!

PR #39, opened 2013-10-16
Fixed link to submission instructions.

zwass:
The link to the submission instructions in `readme.md` was incorrect. It is now fixed.


jlord:
OOoo, good catch! Thanks @zwass!

PR #40, opened 2013-10-16
Add section for International projects

nschonni:
May have some overlap with https://github.com/opengovplatform that is currently under the US section

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -87,6 +87,9 @@ organizations:
   Estonia:
     - vvk-ehk
 
+  International
+    - ogpl
+
   U.S. States:
     - agrc
     - nysenate


afeld:
Interesting... from a quick glance, it seems that http://ogpl.gov.in/ and http://opengovplatform.org/ are the same site under different domains?

afeld:
/cc a couple random people from their contributor pages who might be able to give more insight on the distinction: @pjackson28 @subhransusekhar

I don't think having duplicates in the list is helpful, though it seems like the repositories are separate.  Probably good feedback that the relationship between the two organizations is confusing, though :wink:

pjackson28:
https://github.com/opengovplatform and https://github.com/ogpl are associated but different. Essentially https://github.com/ogpl is seeking to bridge the various projects associated to OGPL (Open Government Platform) whereas https://github.com/opengovplatform is one of the associated projects.

As for http://ogpl.gov.in and http://opengovplatform.org, I believe they are the same site with different domains. Much like with https://github.com/opengovplatform, http://opengovplatform.org/ and http://ogpl.gov.in are associated to https://github.com/ogpl but are different as well (they predate https://github.com/ogpl which will have http://ogpl.github.io for its site).

afeld:
@pjackson28 Thanks for the explanation! I'm still a bit confused about the relationships of the various sites/orgs, but that's beside the point :wink:

_config.yml
```diff
@@ -87,6 +87,9 @@ organizations:
   Estonia:
     - vvk-ehk
 
+  International
```
benbalter, line 90: Just a heads up, valid YAML dictates a trailing `:` here. Fixed in 3dee3d1.

nschonni, line 90: Thanks, that's what I get for not going to bed at a human hour :frowning:

PR #43, opened 2013-10-16
Update _config.yml

leebrian:
Added org name for cdc public health informatics lab.


leebrian:
Adding entry for us-cdc's public health informatics lab.

jlord:
Woah, super rad! Thanks @leebrian!

PR #46, opened 2013-10-16
Add mySociety and OpenAustralia Foundation to Civic Hackers

mlandauer:


jlord:
Yay! Thanks @mlandauer!

PR #47, opened 2013-10-16
Add Team POPONG to Civic Hackers

e9t:
[Team POPONG](http://en.popong.com) is a civic political organization oriented in South Korea.


jlord:
Cool, thanks @e9t!

PR #48, opened 2013-10-16
Add Travis build status to readme

afeld:
[Rendered](https://github.com/github/government.github.com/tree/travis-badge#readme).  This is partially to check that the Travis webhook is configured correctly.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,4 +3,4 @@ script: "./script/cibuild"
 branches:
   only:
     - gh-pages
-    - /*/
+    - /.*/
--- a/readme.md
+++ b/readme.md
@@ -1,4 +1,4 @@
-# GitHub and Government
+# GitHub and Government [![Build Status](https://travis-ci.org/github/government.github.com.png?branch=gh-pages)](https://travis-ci.org/github/government.github.com)
 
 ![screenshot](https://f.cloud.github.com/assets/282759/1335911/28b3ec66-35c0-11e3-8ee6-66272b9f4186.png)
 


afeld:
Well, [the build ran](https://travis-ci.org/github/government.github.com/builds/12605131), but commit status isn't displaying, and the webhook settings are showing an "unknown error".  Investigating.

nschonni:
By default Travis won't build for gh-pages branches. Since this is a *.github.com repo, why not use master as the default branch?

nschonni:
I'm not sure if that will fix it :frowning: 
I guess it would actually need to move to government/government.github.com to be able to use the mater branch for pages though.

afeld:
@nschonni The builds [are being triggered](https://travis-ci.org/github/government.github.com/builds/12607077), so I think it's a different problem.  We re-created this repo right before open-sourcing it, which I'm guessing caused a problem w/ Travis talking to the [commit status API](http://developer.github.com/v3/repos/statuses/#create-a-status).  We'll get it all straightened out.

nschonni:
Ah, you're right. From the docs

> Note that the gh-pages branch will not be built unless you add it to the whitelist (branches.only).

Sometimes the Travis hook needs to be removed and re-enabled if a repo is renamed or transferred. You should be able to do it if you have admin access to this repo.

afeld:
Hmm, now it seems to be working?

screen shot 2013-10-18 at 7 55 26 pm

afeld:
Build passed!  And it told us!

PR #56, opened 2013-10-16
Rename /stories/sunlight to /stories/modern-approach-to-open-data

konklone:
This is the same PR as #23 (I didn't have permissions to re-open it), but with a redirect at /stories/sunlight. It uses a 0 (immediate) timer to do the redirect, which [search engines treat as a 301](http://sebastians-pamphlets.com/google-and-yahoo-treat-undelayed-meta-refresh-as-301-redirect/). It also contains a `canonical` link href to the new post, in case that helps.

I believe Github Pages blocks all plugins, otherwise I'd have used [jekyll-pageless-redirects](https://github.com/nquinlan/jekyll-pageless-redirects/) to implement it.


benbalter:
@konklone thanks for putting this together. Only hesitation is that it adds a post which would then appear twice in index / archive. I'm gonna cherry pick  4fc28df for now to get the rename in there, and will leave this open until we can implement a better redirect, perhaps even client side via the 404 via #65.

konklone:
Just sorting out the threads - I think you :+1:'d this?

konklone:
I had a syntax error leftover from a port from a Jekyll plugin - fixed it just now, PR is updated.

benbalter:
I think we're good on this front? http://government.github.com/stories/sunlight should now direct users to the proper post (via automagic google search).

konklone:
I get a 404 on that URL now. If you replicated this PR on your own, I think you'll want to add a story file back in for 2013-10-01-sunlight.md, with no layout specified, that contains the meta-redirect HTML.

benbalter:
>  Only hesitation is that it adds a post which would then appear twice in index / archive.

Was hoping this approach would make the content easily findable without cluttering the archive.

konklone:
If it's good for you, it's good for me.

PR #58, opened 2013-10-16
Fix Reversed A/B references for Collaborate Section

chrismeller:
In the Collaborate section, the references to A and B on the associated example were reversed.


benbalter:
Great catch. Thanks for taking the time to package the pull request. :beers:

PR #72, opened 2013-10-16
Fixes footer link to main github.com site

lode:
I guessed the link should point to the main Github website, and not to the status page. If not, than the link text would have to be changed to reflect that.


afeld:
Thanks!

PR #73, opened 2013-10-16
Added hasadna to civic hackers

nonZero:
see more here: http://github.com/hasadna
http://www.hasadna.org.il


jlord:
Thanks @nonZero!

PR #77, opened 2013-10-16
Remove Google fonts

bleikamp:
The fonts are no longer in use in the CSS so we should delete them from the mark up.


jlord:
Whhooops! Thanks @bleikamp!

PR #79, opened 2013-10-16
Add opennorth to Civic Hackers

jpmckinney:
Some of the other civic hackers (e.g. sunlightlabs, mysociety) rely on our code and [data specifications](http://popoloproject.com/) in their projects – I think that's enough to qualify. [Open North](http://opennorth.ca/) is Canadian.


jlord:
Yay! Thanks @jpmckinney!

PR #82, opened 2013-10-16
Adds CodeandoMexico to org list

juanpabloe:


jlord:
This looks so awesome @juanpabloe! But since it's not a governmental agency, we should probably put it in the Civic Hackers part of the list. It may turn out that we need to break the Civic Hackers up into countries, too. We're learning there are so many people doing awesome stuff -- which is great!

juanpabloe:
@jlord yep! we fit better actually on the Civic Hackers category. If you break it into countries in the feature, you'll know where we are from :)
Thanks!

jlord:
Thanks @juanpabloe!

PR #85, opened 2013-10-17
Normalize spacing

nschonni:
- Add [EditorConfig](http://editorconfig.org) file for supported browsers 
- Replace all tabs with soft double spaces


benbalter:
Ha. That's awesome. Realize it's the "less glamorous" side of open source, but thanks for taking the time to package this together. Glad someone's on the case!

PR #89, opened 2013-10-17
To add "infoculture" as civic hacker

ivbeg:
Infoculture ("Informational culture") is Russian NGO dedicated to open government, opendata and open source.


jlord:
Thanks @ivbeg!

PR #95, opened 2013-10-17
Add codeforeurope to _config.yml.

pmackay:


benbalter:
@pmackay it looks like @codeforeuropse is a private organization, not a government entity, correct? Perhaps it should go under "civic hackers".

pmackay:
It's a consortium that works with local government. In that case, shall I create a new pull request for that?

jlord:
You can just add another commit on this PR @pmackay. We're actively trying to figure out the best way to sort all of these, but for now, I think like Code for America, it should be in Civic Hackers since the group, I'm assuming like CfA, works with multiple municipalities on multiple projects in parallel. A little different than the actual governmental entity itself open sourcing its work directly.

pmackay:
Moved codeforeurope to civic hackers section.

jlord:
Thanks @pmackay!

PR #97, opened 2013-10-17
Added gbv (http://www.gbv.de) library network

nichtich:
GBV is Germany's largest network of university libraries.


jlord:
Thanks @nichtich!

PR #98, opened 2013-10-17
Update _config.yml

martent:
Added malmostad (City of Malmö), a Swedish municipality.


jlord:
Wow, awesome @jesperbylund! Thank you!

PR #99, opened 2013-10-17
Create 2013-10-17-xas-software

bruceravel:
Create a _post for the Demeter package of software for XAS data analysis and processing, a project from a physicist at the United States' National Institute of Standards and Technology.

### DIFF
--- a/_posts/2013-10-17-xas-software.md
+++ b/_posts/2013-10-17-xas-software.md
@@ -38,7 +38,7 @@ collaborators, including [Feff](http://www.feffproject.org/),
 [Larch](https://github.com/xraypy/xraylarch).
 
 As a system for linking the capabilities of the backend scientific
-tool to useful and expressive user interfaces,
+tools to useful and expressive user interfaces,
 [Demeter](https://github.com/bruceravel/demeter) allows scientists to
 use data measured at synchrotron facilities to the fullest extent,
 enabling significant advances in many of the scientific areas most
--- a/_posts/2013-10-17-xas-software.md
+++ b/_posts/2013-10-17-xas-software.md
@@ -1,16 +1,47 @@
-[Demeter](https://github.com/bruceravel/demeter) is scientific software for the processing and analysis of a measurement technique called X-ray Absorption Spectroscopy (XAS).  This is a technique familiar to users of the Department of Energy's synchrotron radiation facilities 
+[Demeter](https://github.com/bruceravel/demeter) is scientific
+software for the processing and analysis of a measurement technique
+called X-ray Absorption Spectroscopy (XAS).  This is a technique
+familiar to users of the Department of Energy's synchrotron radiation
+facilities
 
   * [The Advanced Photon Source](http://www.aps.anl.gov)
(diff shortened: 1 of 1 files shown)


benbalter:
@bruceravel really, really cool post (at least the part that I understood! :smile:!) and looking forward to working with you to get it published.

Have you had a chance to look at [the style guide](http://government.github.com/submit/style-guide/)? It was a bit hidden, so you may have missed it. Specifically, I'd love to focus our efforts on two things:
1. Keeping in third-person
2. Providing more context to why your approach was so innovative (which it is). What is the approach normally taken? How was this different? What was the result? 

The technical details are extremely cool, but for a non-brainiac crowd, which I'd imagine most of a readers are, providing more context into elements from your story, which may be similar to their current challenges may resonate a bit more.

bruceravel:
@benbalter : Thanks for the comment.  I was becoming a bit concerned that government.github.com may not have been interested in scientific content.  So it is nice to get some feedback.

I have not, in fact, seen the style guide.  It was, apparently, sufficiently hidden to be missed by me. :smile: 

In the next few days I will find a few moments to edit my entry in light of the style guide and your two specific suggestions.  As I do so, I will be very interested to get your feedback on content as well as form.  I actually worry quite a bit about communicating science effectively to a non-scientist audience (I don't much care for your term -- I doubt I am any smarter than you.  The main difference is that I probably liked my physics class in high school more than you did.)  I am pretty certain that what I do can be made compelling for a broad audience, but feedback will help.

bruceravel:
@benbalter : again, thanks for your feedback.  I put a bit of effort into my text, trying to follow more closely to the style guide and trying to make it a bit more readable for someone not like me.  I'd be grateful for some more feedback at this point.  Cheers!

benbalter:
@bruceravel This is great, and much more "human readable" :smile:. I'd love to see if we can't get it posted this week some time.

It's your post, but one suggestion I might have would be to move the paragraph that begins "scientists from the" and the two paragraphs that follow (so in other words, the last three paragraphs) up above the first two. That way you can ease non-technical users into who you are and what the tool is, before jumping directly into the details of the process.

Either way, an awesome story and looking forward to helping to ensure it's told. Thank you for taking the time to put this together.

/ccing @github/science which is much smarter than me on this stuff and my be able to speak more intelligently on the subject.

benbalter:
:heart: it.

The only thing left on my list would be if you had an image that we could associate with the post (see [government.github.com](http://government.github.com) for some examples from the other post. If you just drag and drop it into a comment here, I'll take care of getting things posted.

bruceravel:
I asked a few of my science-y friends to give it a read and offer some feedback.  We'll see what they suggest.

I will think about a figure.  I don't have just the right thing for this context immediately handy, but I will work on it.

bruceravel:
Art is not my strong suit, but here is what I came up with as an image.

logo_try2

benbalter:
:heart: it.

[Posted at government.github.com/stories/xas-software](http://government.github.com/stories/xas-software/). Thanks again for taking the time to submit the pull request!

PR #104, opened 2013-10-17
Added data.gov.uk

davidread:


benbalter:
@davidread is @datagovuk the official government-run account?

davidread:
Yes it is!

jlord:
@davidread we try and verify accounts with emails which is where the confusion comes from. But I'm convinced  - the data.gov.uk site links to this GitHub org :thumbsup:

PR #105, opened 2013-10-17
Add required quotes around Google Analytics Property ID

kneath:
This should fix Google Analytics reporting, right now it's causing Javascript errors on every page.


benbalter:
Whoooops. Good thing there's no other javascript on the site? Thanks for catching that.

:chart_with_upwards_trend:

PR #106, opened 2013-10-17
Add HSLdevcom to _config.yml

tuukka:


benbalter:
Thanks for the pull request. I believe (for now) it may be better suited under "civic hackers" as it doesn't appear to be an official government account.

tuukka:
Just to clarify, the HSL developer community is run by a public organisation of the regional governments, the Helsinki Region Transport Authority HSL. I understand that the case is not clear cut as most of the participants are not civil servants and we want to keep the community open, but almost all of the current commits under the Github org have been done under a contractual relationship with the government.

In any case, I appreciate if you make the call regarding the classification, and we would be happy under "civic hackers" as well :-)

jlord:
That's a pretty cool group @tuukka, thanks for submitting!

PR #107, opened 2013-10-17
Update _config.yml

palmerj:
Added Land Information New Zealand (LINZ)


benbalter:
Believe @linz may be the runner up for best government username behind @chaos. Thanks for the pull request!

Issue #101, opened 2013-10-17
Add metadata template to Contributing.md


nschonni:
Should include the recommended/required YAML section for the posts

benbalter:
Great idea. Want to take a pass at it?

benbalter:
#111 fixes. Thanks @nschonni!

PR #111, opened 2013-10-17
Add Prose default metadata

nschonni:
Related to #101, but the submit.md wording needs to be decided on.


nschonni:
Oops, need to fix up the label text. Will reopen

nschonni:
Attached is how the new metadata modal looks with the explicit meta data fields
prose post metadata pop up

benbalter:
Oh man. Much excellent. Should have done this originally. Thanks @nschonni

nschonni:
No problem. The only catch right now is that the image upload has to be done using the story editor part (but it goes in the correct folder now).

PR #114, opened 2013-10-18
Add ODI to civic hackers list

Floppy:


jlord:
Most definitely! Thanks @Floppy!

PR #115, opened 2013-10-18
add Chinese city

zhangchunsheng:
add Chinese city named xuzhou


jlord:
Thanks @zhangchunsheng! Do you know if this is officially the city's account or a local organization? We like to have verifiable emails for official accounts.

zhangchunsheng:
Ok,I know.

PR #118, opened 2013-10-19
Adding Honolulu

matthewbadeau:
Though, it's the City and County of Honolulu, I added it to the city section.


jlord:
Awesome! Thanks @matthewbadeau!

PR #119, opened 2013-10-19
update for City of Asheville

daveism:


cobyism:
:metal: :sparkles: Thanks! Also, nice to meet someone else with a `-ism` handle :wink:

PR #123, opened 2013-10-20
Added OKFN Egypt to _config.yml

gr33ndata:


jlord:
I didn't realize the OKFN family was so large, that's awesome! Thanks @gr33ndata!

PR #124, opened 2013-10-20
Update _config.yml

johnolinda:
Added Newark, DE to the list.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -138,6 +138,7 @@ organizations:
     - openlexington
     - CORaleigh
     - Honolulu
+    - Newark
 
   Civic Hackers:
     - infoculture


jlord:
Thanks for submitting @jackivan88! The account doesn't appear to be an official city account though? www.github.com/Newark.

johnolinda:
No, it's not an official city account, will that prevent it being added?

jlord:
Not at all, we're just trying to keep all non-official government accounts in the Civic Hacker category. You've proposed this under US City, so that's why I inquired to see if it was the city's official account.

The account also does not appear to have very much activity, however, and I would like to see this list be comprised of users and organizations active in this space. Perhaps resubmit this when it becomes more active?

Thanks so much for taking the time to contribute!

PR #125, opened 2013-10-21
Update _config.yml

leveille:
National Priorities Project (http://nationalpriorities.org) is an organization whose primary goal is to make the US Federal Budget transparent and accessible.


jlord:
Awesome! Thanks @leveille!

PR #127, opened 2013-10-21
Gh pages valid

XhmikosR:
- Fixed validation errors
- Many css tweaks


assets/css/style.css
```diff
@@ -298,6 +289,7 @@ header:after {
 }
 
 header .logo {
+  background: url("../img/github-gov-black.png") no-repeat 0 0/contain;
```
nschonni: `0/contain` -> `0 contain`

assets/css/style.css
```diff
@@ -471,10 +466,14 @@ img.avatar {
 }
 
 .contribute {
-  background: #FAFAFA;
-  border: 1px solid #EEE;
+  background-color: #fafafa;
+  border: 1px solid #eee;
+  -webkit-border-radius: 4px;
+  -moz-border-radius: 4px;
```
nschonni, line 410: border radius hasn't needed vendor prefixing since FF4 https://developer.mozilla.org/en-US/docs/Web/CSS/border-radius#Browser_compatibility

XhmikosR, line 410: Sure, but it doesn't really cost anything. I'd rather support old browsers when it's so simple.

assets/css/style.css
XhmikosR: Nope, that's wrong.

`background: url("../img/github-gov-black.png") no-repeat 0 contain;` won't work

But `background: url("../img/github-gov-black.png") no-repeat 0 0/contain;` will. The same will with one zero too, but I wanted to be explicit in this case.

nschonni: My point was about the `/` unless it's something like the `/9` IE hack

assets/css/style.css
nschonni, line 410: Your choice, but even Mozilla doesn't support anything below 10 now

assets/css/style.css
XhmikosR: Well the results above are with the latest FF Aurora and Chrome Dev.

assets/css/style.css
XhmikosR, line 410: If it was my choice I wouldn't care about old browsers including < IE 9... But people unfortunately use old crap all the time. And this is a simple fix...

benbalter:
@XhmikosR do you mind merging in `gh-pages` and resolving any conflicts? After that, if the build is :green_heart: would love to get this merged in. Thanks again for your help.

XhmikosR:
I have already rebased this branch against the latest upstream/gh-pages; Travis tests pass  https://travis-ci.org/github/government.github.com/builds/13078236

XhmikosR:
Oh wait, I'll rebase again...

XhmikosR:
Done

benbalter:
Awesome. Thanks again for taking the time to make all these tweaks.

XhmikosR:
No problem.

Now if we could reduce the number of http requests and redirects, that would have a great impact.

The first thing would be not to include all organizations in index.html; they are hidden anyway with `overflow: hidden;` in `#the-orgs`.

XhmikosR:
For reference, http://gtmetrix.com/compare/rH3HRVRA/LGXbwqp1

As you can see there are so many requests.

PR #128, opened 2013-10-21
Gh pages images

XhmikosR:
- remove unused images from the repository
- use a 16x16 favicon instead of 32x32
- add external images to the repository
- losslessly compress the images to save a lot of bytes

```
Before: 19 images 4,30 MB (4.514.694 bytes)
After:  20 images 1,50 MB (1.574.465 bytes)
```


benbalter:
Once again, a great contribution. Thanks for taking the time to clean these up.

PR #129, opened 2013-10-21
Update Readme.md, _config.yml and clean up package.json

XhmikosR:
PS. If any PR's conflict, just ping me and I'll update them.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,2 @@
-_site
 .DS_Store
-node_modules
+/_site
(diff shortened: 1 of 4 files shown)


benbalter:
This is great, thanks. One thing though, `vendor` needs to be in the exclude list... see the failed build status.

XhmikosR:
Totally missed it because Travis seems to fail for the upstream branch too.

PR rebased and updated.

benbalter:
Odd. Not sure why those tests are suddenly failing. WIll investigate elsewhere. Thanks again!

PR #130, opened 2013-10-21
added CodeForTomorrow to _config.yml

schee:
Code for Tomorrow Foundation Initiative was founded in 2012 and serves as one of the most important organizations on open data and open government from Taiwan.


jlord:
Awesome, thanks for adding @schee!

PR #132, opened 2013-10-21
Update _config.yml

parfenov:
OKF Ireland

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -1,8 +1,7 @@
 #list of GitHub organization usernames, grouped by type
 organizations:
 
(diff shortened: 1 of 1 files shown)


nschonni:
Should probably be under the "Civic Hackers" section like the rest of the OKF\* organizations

jlord:
Yep, @parfenov will you put it under Civic Hackers since it is not an official government account? Thanks for submitting!

parfenov:
I think it's done.

jlord:
Thanks @parfenov!

Issue #75, opened 2013-10-16
How should the levels of government be laid out


nschonni:
Not to promote silos :trollface: but ...
There is currently a slight disconnect with US splitting between Federal, State, county, and City vs. the rest of the world.
1. Each country splitting between levels (with separate h2's and grids?)
   - Country/Federal
     - Country A
   - State/Province
     - Washington
     - Ontario
   - City/County
     - ...
2. Country as the high level
   - US
     - Federal
       - ...
     - State

afeld:
While it's good to recognize this structure, I don't know that it covers the full range, either.  See #21.

I imagine that, down the road, it would be nice to have some sort of tagging/filtering system for level, location, and maybe even focus of the organization.

nschonni:
No, it's definitely not an exhaustive list, but just 2 possible top level headings. The civic hacker folks may want some distinct levels too ...
Maybe use something like DataTables for a sortable view http://wet-boew.github.io/wet-boew/demos/tables/tables-en.html#inisort

jlord:
I'm definitely learning it's a lot more complex than how we organized it originally. Agree @nschonni that we may end up needing to make distinctions between civic hackers, too. 

Let's keep watching as groups trickle in and definitely dream up a better way to showcase all the orgs.

benbalter:
Now's definitely the right time to start this discussion, although I'd think we'd want to hold off a bit before making any changes. The question I keep asking is "what are we trying to solve for?"
- Help like-minded organizations to find each other
- Help citizens to quickly and easily find organizations relevant to their location / interests
- Make an extremely powerful graphic / statement about the strength of open source in government
- Keep it as simple as possible for non-technical contributors to add their name to the list
- (and don't want to promote silos as @nschonni pointed out)

I was thinking initially that top level should be country, and subdivide from there, but spaces + YML is going to be a big headache for the uninitiated.

Only reason US is split up like it is, is just because of size, which I immagine civic hackers may soon need to be split as well.

smathermather:
Added Special District in order to add Cleveland Metroparks, which I know isn't common, but we didn't fit in any other category.  "Local" might be a more flexible category.

nschonni:
:bulb: like #101 we could add a folder with specific frontmatter

``` yaml
prose:
  media: "images"
  metadata:
    _orgs:
      - name: "layout"
        field:
          element: "hidden"
          value: "org"
      - name: "github"
        field:
          element: "text"
          label: "GitHub org"
          value: ""
[...]

PR #133, opened 2013-10-21
Fixes sorting

leucos:
Some countries were not properly sorted. 
Moved E.U. to the bottom, near non-country organizations

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -1,6 +1,71 @@
 #list of GitHub organization usernames, grouped by type
 organizations:
 
+  Australia:
+    - actgov
+
+  Argentina:
+    - gcba
+
+  Belgium:
+    - onroerenderfgoed
+
+  Brazil:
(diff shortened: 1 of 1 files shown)


jlord:
Nice, thanks @leucos!

benbalter:
Are we sure that an alphabetical country listing is the right way to go here? Visually, it's a lot less stunning because many of the A's only have one or two accounts:

screen shot 2013-10-21 at 12 55 14 pm

screen shot 2013-10-21 at 12 57 04 pm

/cc #75

leucos:
Mmmh, yes that's a point. It could be nicer (and more interesting) to move up entries depending on how many organization they contain ?

nschonni:
@benbalter the flip to that is that it now highlights that (at least) 13 countries that have repos on the page load instead of the 5 when the US block is first

PR #139, opened 2013-10-23
[WIP] Add organization metadata

nschonni:
## Do not merge

Sample of moving organizations from _config.yml to a separate markdown format

org metadata on prose

This would allow people to write a brief description of the organization and provide the required metadata to categorize them on the Community page.

Related to gh-75

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -298,3 +298,49 @@ prose:
               value: "open data"
             - name: "Open Source"
               value: "open source"
+
+    _orgs:
+      - name: "layout"
+        field:
+          element: "hidden"
+          value: "org"
+      - name: "github"
+        field:
+          element: "text"
+          label: "GitHub org"
+          value: ""
+          type: "text"
+      - name: "fullname"
+        field:
+          element: "text"
+          label: "Full Name"
+          value: ""
+          type: "text"
+      - name: "description"
+        field:
+          element: "textarea"
+          label: "Description"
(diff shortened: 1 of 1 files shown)


benbalter:
Thanks for taking a pass at this. I'm a little skeptical about an approach that goes down this path. A few reasons:
- If this is the direction, I'd love to iterate there, e.g., adding the metadata to _config.yml, etc.
- There's significant value in having contributors use the web flow to submit a pull request
- This adds considerable complexity
- Right now, we don't expose most of that information to the public

nschonni:
> If this is the direction, I'd love to iterate there, e.g., adding the metadata to _config.yml, etc.

Not sure I understand this point? Are you talking about continuing to add metadate under the `organization` heading int the `_config.yml`?

> Right now, we don't expose most of that information to the public

The only piece not being collected/already on a GH profile is "Description" which should actually be removed in favour of the markdown body.

> There's significant value in having contributors use the web flow to submit a pull request
> This adds considerable complexity

Kind of, it more enforces the prose.io workflow over raw YAML editing in the ACE editor.

benbalter:
Going to close this as stale, since we're no longer using Prose or `_config.yml` to store orgs.

@nschonni greatly appreciate all your contributions to date, and hope we can figure out a way to incorporate some of this additional metadata programmatically, so we're not pushing the burden onto contributors.

PR #140, opened 2013-10-23
Update _config.yml

ghost:
Added French National Security Authority (ANSSI-FR)

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -80,6 +80,7 @@ organizations:
     - erasme
     - etalab
     - VilleDeMontpellier
+    - ANSSI-FR
 
   Germany:
     - gbv


benbalter:
@ANSSI-FR, you may wish to consider [converting your user account to an organization account](https://help.github.com/articles/converting-a-user-into-an-organization). Most if not all other government organizations on GitHub are organizations. See [the best practices guide](https://government-community.githubapp.com/government/best-practices/blob/master/docs/account-management.md#organization-or-user) for more information, but in short:

> ## Organization or user
> 
> Agencies should be represented by [GitHub organizations](https://help.github.com/articles/what-s-the-difference-between-user-and-organization-accounts#organizations). Think of these as roughly analogous to a Facebook Page.  Organizations do not have their own login, but are administered by users. Organizations have their own, distinct online presence, can own repositories, and can have unlimited members and teams.

jlord:
Going to merge this in but ditto on turning the account into an organization instead of a user so that staff can contribute more easily! Thanks @ANSSI-FR!

PR #146, opened 2013-10-24
Fixed typo

pborreli:


benbalter:
Ha. Great catch. Thanks for that, as always @pborreli!

PR #154, opened 2013-10-26
Move inline style to style.css.

XhmikosR:
Also remove the unused .post-image class.


XhmikosR:
@benbalter: Did a few more improvements, let me know if you need more info.

benbalter:
Awesome. And thanks for making them granular. :smile:

PR #160, opened 2013-10-29
Update _config.yml

coomsie:
Added CanterburyRegionalCouncil a regional govt agency of New Zealand


jlord:
Cool, thanks @coomsie!

PR #166, opened 2013-10-31
Update _config.yml

breyten:
OpenState is a foundation working to improve open data and transparency in The Netherlands and The European Union. We regularly organize hackathons as well. Our website is at http://openstate.eu/


jlord:
Awesome! Thank you @breyten!

PR #169, opened 2013-11-01
Be consistent with the quotes type.

XhmikosR:
- use single quotes inside double quotes
- remove unneeded quotes from posts' descriptions

This change is just cosmetic, since jekyll works fine either way but it's better not to complicate things.


benbalter:
Thanks!

PR #170, opened 2013-11-04
Add CMAP-REPOS to _config.yml

nmpeterson:
Added CMAP (Chicago Metropolitan Agency for Planning) to U.S. Special Districts


jlord:
This makes the former-urban-designer-me super excited to see! Thanks @nmpeterson!

PR #173, opened 2013-11-05
Reduce http requests

XhmikosR:
I also made them responsive.

### DIFF
--- a/index.html
+++ b/index.html
@@ -50,13 +50,18 @@ title: GitHub and Government
   <div class="container">
     <div id="orgs">
       <div class="the-orgs">
+    {% assign count = "0" %}
     {% for type_hash in site.organizations %}
       {% for org in type_hash[1] %}
+        {% capture diff %}{{ count | minus:60 }}{% endcapture %}
+        {% if diff contains '-' %}
         <div class="organization">
           <a href="https://github.com/{{ org }}" title="{{ org }}">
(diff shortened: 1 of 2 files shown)


XhmikosR:
Before: 336 requests 1,7 MB


---

After: 107 requests 1,3 MB

index.html
```diff
@@ -48,19 +48,24 @@ <h1 id="join-the-revolution">Join the Revolution</h1>
...
 
   <div class="container">
-    <div id="orgs">
-      <div class="the-orgs">
+    <div class="row">
+      <div class="span12">
+    {% assign count = "0" %}
     {% for type_hash in site.organizations %}
       {% for org in type_hash[1] %}
+        {% capture diff %}{{ count | minus:44 }}{% endcapture %}
```
benbalter, line 56: What exactly are you trying to do here?

benbalter:
I believe part of the initial design vision was to have the orgs overflowing out of the box to indicate that there were many, many more. Can still limit, but should most likely be expanded by one additional row in each direction to maintain that effect.

index.html
XhmikosR, line 56: 44 is the number of organizations shown; 4x11

XhmikosR:
Well, currently that row isn't responsive at all. Personally, I prefer this way, but feel free to improve it after merging this.

PR #175, opened 2013-11-05
Limit http requests on the home page

### LINKED ISSUE #151: Reduce http requests and redirects
Initially mentioned here  https://github.com/github/government.github.com/pull/127#issuecomment-27144625.

There are so many http requests ~320. And most of them are because of redirects from github to gravatar.

Now since most organizations in `#orgs` are hidden with `overflow: hidden;`, they shouldn't be in index.html in the first place IMHO.

Another approach would be to use a css sprite but that won't be so flexible.

benbalter:
Limit the number of HTTP requests on the index page org logo view to the number visible, which may slow things down, especially as the list grows. This is similar to and influenced by @XhmikosR's efforts in #173 with a few notable changes:
1. Doesn't affect responsiveness 
2. Doesn't affect the visual output
3. I believe the logic of what's actually happening is more immediately apparent 
4. Move the limiting factor to page YML frontmatter for easier changability

@XhmikosR's if this pull request is merged in favor of #173, perhaps it may make sense to open up individual pull requests for 1 and 2 above so that we can give those changes the proper discussion they deserve?

screen shot 2013-11-05 at 12 19 54 pm

Fixes #151.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,6 +1,7 @@
 ---
 layout: home
 title: GitHub and Government
+org_count: 60
 ---
 
 <div class="stories container">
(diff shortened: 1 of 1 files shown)


benbalter:
@XhmikosR I believe so, here's what happens when I change `org_count` to 10.

screen shot 2013-11-05 at 12 20 32 pm

XhmikosR:
OK, looks simpler then.

Personally, I don't find this approach good with the logos being cut off. But it's your call.

The only thing left after this is responsiveness.

benbalter:
> Personally, I don't find this approach good with the logos being cut off. But it's your call.

It's less so that one approach is better than the other, it's that it's a big enough change, or an unrelated enough change that it deserves its own pull request so that it can be surfaced to the community for broader discussion, rather than being snuck into a tangentially related change just because the two are physically near each other in the code.

XhmikosR:
Well I can make a PR after this is merged for that part.

But generally, I like this approach with limiting the orgs. When I was working on it, I was getting errors when I compared with less/greater than. But since it works, it's simpler than my approach, so go for it.

benbalter:
>  I was getting errors when I compared with less/greater than. 

If I had to guess, it was because you were using assign, which would treat `count` as a string, not a variable.

But either way, thank you again for all your contributions. :sparkling_heart: :smile:

XhmikosR:
That was my thought too, but couldn't fix it so I worked around it.

No problem, I'm learning new stuff along the way. :)

PR #179, opened 2013-11-06
Move organizations to a datafile

parkr:
Contingent on GH Pages moving up to Jekyll v1.3.0 or greater.

PR #180, opened 2013-11-06
Move organizations to data file

troyswanson:
Why wait? Let's take advantage of Jekyll 1.3.0's new data structure!

One thing I'm not sure about is the `script/fetch_us` file. This seems like a Ruby script that I don't know anything about. :cry:


troyswanson:
Looks like @parkr beat me by a few minutes! Closing in lieu of #179.

PR #182, opened 2013-11-06
CSS cleanup and fixes

XhmikosR:
Overview:
1. Fixed footer links with low resolutions; now the footer links don't overlap anymore
2. Fixed `:not` selector
3. Simplified media queries
4. Removed zooming from homepage-image
5. Removed many unused selectors

### DIFF
--- a/assets/css/icons.css
+++ b/assets/css/icons.css
@@ -7,19 +7,16 @@
        url("/assets/fonts/github-octicons.ttf") format("truetype"),
        url("/assets/fonts/github-octicons.svg#github-octicons") format("svg");
   font-weight: normal;
-  font-style: normal;
-}
+  font-style: normal; }
 
 .octicon {
-  font: normal normal 16px "github-octicons";
-  line-height: 1;
+  font: normal normal 16px/1 "github-octicons";
   display: inline-block;
   text-decoration: none;
   -webkit-font-smoothing: antialiased; }
 
 .mega-octicon {
-  font: normal normal 32px "github-octicons";
-  line-height: 1;
+  font: normal normal 32px/1 "github-octicons";
   display: inline-block;
   text-decoration: none;
   -webkit-font-smoothing: antialiased; }
--- a/assets/css/style.css
+++ b/assets/css/style.css
@@ -449,18 +449,18 @@ a.open.source:hover { color: #00acef; border-color: #00acef; }
 /* Media Queries */
 /* ----------------------------------------------------------------------------------- */
 
-@media only screen and (max-width: 979px)  {
+@media (max-width: 979px) {
   .no-mobile { display: none !important; }
 }
 
-@media only screen and (min-width: 650px) and (max-width: 767px) {
+@media (min-width: 650px) and (max-width: 767px) {
   .span4, .span8, .span10, .span12 { width: 100%; }
   nav { float: left; padding-top: 20px; width: 428px; text-align: left; }
-  header nav a:first-of-type { margin-left: 0;}
+  header nav a:first-of-type { margin-left: 0; }
   img { width: 100%; }
 }
 
-@media only screen and (min-width: 320px) and (max-width: 649px) {
+@media (min-width: 320px) and (max-width: 649px) {
   .span4, .span8, .span10, .span12 { width: 100%; }
   a { word-wrap: break-word;}
   nav { float: left; padding-top: 20px; }
--- a/assets/css/style.css
+++ b/assets/css/style.css
@@ -508,13 +508,12 @@ a.open.source:hover { color: #00acef; border-color: #00acef; }
 /* Media Queries */
 /* ----------------------------------------------------------------------------------- */
 
-@media only screen and (min-width: 768px) and (max-width: 959px)  {
-  li.no-mobile { display: none; }
+@media only screen and (max-width: 979px)  {
+  .no-mobile { display: none !important; }
 }
 
 @media only screen and (min-width: 650px) and (max-width: 767px) {
   .span4, .span8, .span10, .span12 { width: 100%; }
-  li.no-mobile { display: none; }
   nav { float: left; padding-top: 20px; width: 428px; text-align: left; }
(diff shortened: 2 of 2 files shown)


XhmikosR:
_Bump_

This branch has important fixes and should be merged asap...

jlord:
@XhmikosR thanks so much for all your contributions. 

This repository, however, is not the only thing we work on :) As all PRs take time to review, we need to prioritize changes to the site that improve accuracy, representation and usability. 

Opinionated style changes and minor improvements that yield very little noticeable enhancements, I hope you'll understand, cannot be merged on demand and in some cases, needn't be merged (there are many ways to write CSS). PRs that strictly address errors in the site enable energy to be focused on the site's effectiveness in engaging and connecting those in the open government space.

XhmikosR:
This PR is pretty clear though; it only has fixes, not style changes apart from the background-size removal.

jlord:
- adding box-sizing does not greatly improve the usability of the site, is not required and the page was designed without it
- i believe footer links shipped working so I'm guessing this commit fixes an unintended consequence of a previous style PR :+1: 
- incorrect selector name fix :thumbsup: 
- remove unused selectors :+1: 
- removing only screen does not greatly improve the usability or performance of the site
- background image zoom is intentional
- font shorthand property is a semantic preference
- the not selector works both ways; semantic preference 

The majority of commits are style or semantic preferences or minor changes with little noticeable improvements and required a fair amount of time to dive into. I'm closing this PR but welcome one that addresses the three :+1: with corrections to errors.

XhmikosR:
- box sizing makes sure the box model is what we want to. At the moment it's not what it should be
- using screen when no other media is specified is useless
- background image zoom might be intentional but looks awful in low res when the homepage-image takes the full width
- the not selector does **not** work both ways; comma doesn't work
- for font what you have is again useless since you already use the shorthand font property just above. So specifying line-height separately it's just extra bloat

XhmikosR:
Message taken I guess. I'll stop bothering with GitHub's own stuff from now on.

assets/css/icons.css
```diff
@@ -7,19 +7,16 @@
...
        url("/assets/fonts/github-octicons.svg#github-octicons") format("svg");
   font-weight: normal;
-  font-style: normal;
-}
+  font-style: normal; }
 
 .octicon {
-  font: normal normal 16px "github-octicons";
-  line-height: 1;
+  font: normal normal 16px/1 "github-octicons";
```
brntbeer, line 13: as a non-designer, this increases the barrier of entry for me contributing to this project because it's not typical css coding style.

This is less explicit to me, requiring me to go look up how to code to a style similar to this throughout the project.

assets/css/style.css
```diff
@@ -205,38 +178,6 @@ nav {
...
-  border: none;
-}
-
-#top-nav .octicon-mark-github {
-  float: left;
-  width: 24px;
-  height: 25px;
-  font-size: 25px;
-}
-
```
brntbeer, line 239: Love these kind of changes. Remove the cruft that isn't used anymore! :smile:

brntbeer:
Commented on some things!

I think the goal here with the time usage of GitHubbers working on other projects is that there's some changes that are more commonly looked at as opinion changes, and some that are super helpful (the deletion of code for example) and we're just trying to make sure we don't have to look through all of the tiny changes for small payoff is all!

benbalter:
> Message taken I guess. I'll stop bothering with GitHub's own stuff from now on.

@XhmikosR to be clear, your passion and dedication are very much appreciated. 

As was pointed it out in https://github.com/github/choosealicense.com/pull/121 and https://github.com/github/government.github.com/pull/149, pull requests for individual improvements are a lot easier to review than catch-all, generic pull requests with many opinionated changes. 

In this case, for example, had this been individual, feature-driven pull requests, three would have already been merged. Chances are if your pull request body requires multiple bullets summarizing your changes, there's simply too much in there. Granularity FTW! 

There's always a lot to do, and nothing's ever finished or perfect — far from it in fact. At GitHub, as much as possible, we try to focus on front-facing improvements that provide immediate value to users. To borrow a line from the Zen of GitHub, "Practicality beats purity.".

Keep the pull requests coming, but granularity (and patience) is the way to go. :smile:

assets/css/icons.css
XhmikosR, line 13: Well, this is the standard shorthand property which is already being used as you can see. Having the line height separately doesn't make any sense at all in this case...

brntbeer, line 13: it is the standard shorthand, but shorthand raises readability.

XhmikosR, line 13: I disagree. If that's your problem, you wouldn't use any shorthand properties at all. And anyway:
1. There's always a balance between readability/maintenance and performance (see the part of my PR where I remove the useless stuff)
2. I honestly can't think of anyone who works with CSS that finds your way better

XhmikosR:
@benbalter: I understand that, but you need to understand that these changes are made along the way of me fiddling with the source. In this case as you can see, the branch is pretty clear.

Also, I would like to point out the nice feature called cherry picking.

Making a new PR for just a single thing, just doesn't work for me, sorry. I keep the patches grouped. And I don't plan to spend my time to formalities; I'd rather fix stuff like this PR does yet @jlord prefers to just close it.

assets/css/icons.css
skalnik, line 13: I work with CSS and haven't seen this shorthand until today. That being said, I don't think it's a big deal since no one will likely be tinkering with this font property.

XhmikosR, line 13: I dunno, it was one of the first things I learned to be honest. And I like using the shorthand properties. Except for the background-size which doesn't work in old IE.

http://www.w3schools.com/cssref/pr_font_font.asp

> The properties that can be set, are (in order): "font-style font-variant font-weight font-size/line-height font-family"

XhmikosR:
I have made a couple of changes to address the `background-size` removal @jlord mentioned; I removed it for low res only. My css branch is updated.

PR #183, opened 2013-11-06
Fixing instructions for creating pull request

leahbannon:
I think this process is different for public users vs. GitHub employees. cc @benbalter


jlord:
:thumbsup:

benbalter:
Just walked through with a non-staff account, and I believe the problem is a missing step, not a missing feature. Prior to clicking into `_config.yml`, you need to click the branch drop down to create a new feature branch. Once that's done, the rest of the instructions, as written are correct.

@leahbannon if you are able to reproduce, might you be able to update your pull request to correct the instructions?

leahbannon:
Ok, submitted. I'd like to understand branches better - I'd be very grateful if someone could share a good description of how to use branches properly and why they're preferable to the other approach I submitted.

PR #186, opened 2013-11-07
Uncomment users in _config.yml

benbalter:
Users were commented out to prevent errors when the listing relied on the orgs API. since we're static HTML, we can let the non-org listings slide.

Hoping to follow up separately with a script to verify which are users and which are orgs so we can reach out and get them on the straight and narrow. :smile:


jlord:
:+1:

PR #192, opened 2013-11-14
Fix ie8 octicons

### LINKED ISSUE #188: Question about github-octicons
How were github-octicons generated? I ask because the .eot file doesn't work on IE8, but if I regenerate the font files from the .ttf one using [fontsquirrel](http://www.fontsquirrel.com/tools/webfont-generator) it works fine.

cameronmcefee:
Fixes #188, updates fontname mismatch that was breaking octicons in ie8.


XhmikosR:
Works fine, good to go from my side :+1:

benbalter:
Thanks for this.

PR #198, opened 2013-11-20
limit image height on single posts

### LINKED ISSUE #197: Limit image height on single story page
For tall images, would be awesome to dynamically resize like we do on the home page (I think as a centered background with a max-height?)

screen shot 2013-11-20 at 3 31 03 pm

benbalter:
This is pushing the limits of my css-foo, so a second set of :eyes: would be awesome.
## CF

screen shot 2013-11-20 at 3 31 03 pm
## With

screen shot 2013-11-20 at 3 38 24 pm

Fixes #197.

### DIFF
--- a/_layouts/story.html
+++ b/_layouts/story.html
@@ -3,7 +3,7 @@
 <div class="container">
   <div class="row">
     <div class="span8">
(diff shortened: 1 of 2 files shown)


jlord:
:thumbsup:

PR #199, opened 2013-11-20
Resizing ncip image

arfon:
The NCIP image is pretty tall so resizing its height (feel free to ignore!)


benbalter:
Ha. See #198 which tried to do the same thing from the other side. Good call. Thanks.

arfon:
Ha! Missed that.

PR #200, opened 2013-11-20
Fixed typo.

rianrainey:


benbalter:
@rianrainey great catch, and meta too. Thanks!

rianrainey:
@benbalter I thought the irony was great with that too.

I was wondering if you had any suggestions for getting the ball rolling for my city, Nashville, to be a part of this movement? What was the first step for other cities? Can I just create a 'Nashville Github Government' organization and have people submit issues to it as suggested on the [Projects](http://government.github.com/projects/) page? Or do I need a local government official on board first?

I'd like to create a 'Nashville' organization, but it looks like one is already created? Do you know the story/purpose of this [Nashville Repo](https://github.com/Nashville)

benbalter:
> I was wondering if you had any suggestions for getting the ball rolling for my city, Nashville, to be a part of this movement? What was the first step for other cities? Can I just create a 'Nashville Github Government' organization and have people submit issues to it as suggested on the Projects page? Or do I need a local government official on board first?

There's generally two routes you can take, and in many cities, we see both as the first step:

The first way would be to stand up an open source or open data presence dedicated to Nashville outside of government to create a place for other civic hackers to gather and exchange ideas. Reaching out to [the local Code For America brigade](http://www.meetup.com/Nashville-Brigade-Code-for-America/) may be a great start here.

The second way is to reach out to your local government and make the case for why they should establish a presence themselves, either just to get feedback from developers, or even better, to open source code or release data that they've created. Having someone to hold their hand throughout the process goes a long way.

rianrainey:
Thanks @benbalter. I'll look into these.

PR #211, opened 2013-11-25
Adds datauy from Uruguay to organizations.

gabelula:
Website: www.datauy.org

### DIFF
--- a/_data/organizations.yml
+++ b/_data/organizations.yml
@@ -128,6 +128,9 @@ International:
   - ogpl
   - rOpenGov
 
+Uruguay:
+  - datauy
+
 U.S. States:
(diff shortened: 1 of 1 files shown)


benbalter:
@gabelula thanks for the pull request. It looks like @datauy is a non-profit, not an official government organization, is that correct? If so, at least for now, I believe it's better suited in the "civic hackers" group.

gabelula:
Yes. Where do I have to add it then?

benbalter:
In the same file, just under the "civic hackers" heading towards the bottom of the file.

benbalter:
Sorry for the delay! Glad to have you added.

PR #221, opened 2013-12-09
Update libs

XhmikosR:


XhmikosR:
How about removing modernizr and using html5shiv directly? That's what's currently being used from modernizr atm...

PR #223, opened 2013-12-12
Include only the Bootstrap components that are used.

XhmikosR:
This results in 1/4 of the CSS size.


benbalter:
Thanks for the pull request. Even though less performant, I think we're better off not maintaining a custom build of bootstrap in terms of upkeep costs.

XhmikosR:
There's no cost; this is made with the official bootstrap's customizer from their site...

PR #226, opened 2013-12-16
NCIP post - More of a call-to-action voice for forks and pull requests

randomecho:
Between

> We are strongly encouraging forks and pull requests.

and

> We strongly encourage forks and pull requests.

the former sounds tentative, the latter a more definitive call to arms.

Also puts the capital_p_dangit in GitHub Pages.


benbalter:
:heart: the WordPress reference. Thanks for the Pull Request.

PR #227, opened 2013-12-16
NIST blog post missing hyphen in state-of-the-art

randomecho:
From http://government.github.com/stories/xas-software/


afeld:
Thanks!

PR #229, opened 2013-12-17
Improve older browsers support

XhmikosR:
Basically
1. get rid of `background-size: contain`
2. use Bootstrap's `row-fluid`
3. switch to using `img` for the logo

Before:
before

After:
after


_includes/header.html
```diff
@@ -26,7 +26,7 @@
 
   <header class="clearfix">
     <div class="container">
-      <a href="/"><span class="logo">GitHub and Government</span></a>
+      <a href="/"><span class="logo"></span></a>
```
nschonni: Text should be moved to the title attribute at least for screen readers.

XhmikosR: No idea about that since I've never used them and I have no idea how to test them on my desktop, but I guess I'll borrow Bootstrap's `.text-hide`.

https://github.com/twbs/bootstrap/blob/master/dist/css/bootstrap.css#L5558

nschonni: You can just use 

``` html
<a href="/"><span class="logo" title="GitHub and Government"></span></a>
```

But it is more appropriate to use an `` tag here with `alt="GitHub and Government"`

randomecho: Turn off loading images and that gives some approximation to the screen reader experience.  
For example, if you were using Opera 12 you'd Shift + I to toggle images off/on

XhmikosR: @nschonni: I'm not going to fix any accessibility stuff in this PR. You are welcome to fix those issues yourself after this PR is merged.

nschonni: Fair enough, I would revert this line in the PR then.

XhmikosR: Atm I'm trying to fix the navbar elements overlapping in 754px-767px width.

nschonni: Thanks, looks good :+1:

XhmikosR:
@nschonni: now it should be ok.

benbalter:
@jlord this is beyond my comfort level. Any opinion on?:

> 1. get rid of background-size: contain
> 2. use Bootstrap's row-fluid

XhmikosR:
@benbalter: I made the navbar responsive but it's on top of this branch and my other PRs. I want to gather all my responsive related changes in one PR but they conflict with the current PRs...

Edit: here's how it looks like so far. I tried to keep the style close to what is currently is; only needs a little tweaking for older IE. But I'm open to any suggestions.

benbalter:
@XhmikosR just tested this out locally. Wow. This is great. Thanks for taking the time to bundle all this together. Such an improvement! :tada:

XhmikosR:
@benbalter: your bower stuff broke this.

PR #231, opened 2013-12-18
Get rid of modernizr.js and use html5shiv.js directly.

XhmikosR:
Tested with IE8 everything works like before since only the shim was used from modernizr anyway.


nschonni:
The file appears to be minified (good idea), but it isn't reflected in the file name.

XhmikosR:
Because that's the default filename it ships with.  https://github.com/aFarkas/html5shiv/tree/master/dist

XhmikosR:
Also neither modernizr.js had "min" in its filename and it was minified too. This is too much nitpicking even for me :P

benbalter:
Think that's a great idea. Do you mind updating now that https://github.com/github/government.github.com/pull/246 has been merged?

XhmikosR:
I can have a look tomorrow.  Bower is nice and all when you have a way to improve things locally something that isn't implemented here. Otherwise it's another case of bloating things for a little usability...
But anyway that was something for that pr before it was merged.

XhmikosR:
Updated.

PR #233, opened 2013-12-19
typo (removed repetition of acronym in 2nd paragraph)

bruceravel:


jlord:
Good catch, thanks @bruceravel!

PR #234, opened 2013-12-20
"Crowd source" is a stilted way to refer to a group of contributors in "A Modern Approach to Open Data"

randomecho:
"crowd source" is a verb, so using it to label a group of people makes the sentence a little awkward.

"crowd sourced efforts" attempts to deliver the same drive and people but without rewriting the entire sentence fragment in a passive voice.


jlord:
I'm :thumbsup: on these _and_ learned it's The New York Times, not the New York Times :)

Thanks @randomecho!

PR #235, opened 2013-12-21
Fix quote about WET funding sources to match intent, transcript

randomecho:
The verbatim quote is [at 54:34 of the video](http://www.youtube.com/watch?v=zBLxKcLIIWg&t=54m34s)

> It's a... it's community resources [...]

But for the sake of clarity and grammar, should best be written as

> It's community resources [...]


afeld:
Build failure wasn't you're doing, so we can address it separately.  Thanks!

randomecho:
That's a relief. Was wondering why a Markdown file not touching anything else was flagging a build fail.

afeld:
It's using [HTML::Proofer](https://github.com/gjtorikian/html-proofer) to test against live URLs, which unfortunately makes it non-deterministic.  Pros and cons.

benbalter:
@afeld we could wrap HTML Proofer with [VCR](https://github.com/vcr/vcr)?

afeld:
Good question... feels like it would kinda defeat the purpose, though?  Granted, if a URL goes away, then there's not a lot we can do besides remove the link.  I suppose all new links would be caught.

/cc @gjtorikian

gjtorikian:
I'm not against such a configurable option (even though I personally wouldn't use it--consider every PR scanning through links a recheck of an often unchecked section of testing). I don't know how to work with the use case well enough to learn how to work the VCR (heh).

PR #237, opened 2014-01-11
Added LocalGovDigital

PhilRumens:
Added Uk Local Government organisation LocalGov Digital.

### DIFF
--- a/_data/organizations.yml
+++ b/_data/organizations.yml
@@ -77,6 +77,7 @@ United Kingdom:
   - lichfield-district-council
   - ministryofjustice
   - dfid
+  - LocalGovDigital
 
 United States:
   - adlnet


jlord:
Thanks for adding @PhilRumens!

Is @LocalGovDigital an actually governmental agency or a civic consultancy that works with governments? If the latter, I think it makes most sense to put them in the 'Civic Hackers' portion of the file. The file has gotten quite long now, but it's the last section in the file :)

PhilRumens:
Hi @jlord 

We're a collection of UK local government employees working together as the organisation LocalGov Digital. We're employed by different agencies but work together on common goals and aims. Our website is http://sites.idea.gov.uk/localgovdigital/ which you can see is a .gov address.

So we're not a governmental agency or a civic consultancy, but I think we belong under the UK section as we're all employed by local government agencies, doing work for local government.

I hope this helps but please let me know if you need any more info.

jlord:
Sounds good to me, :thumbsup:  thanks @PhilRumens!

PR #239, opened 2014-01-15
Update organizations.yml

jungshadow:
Adding votinginfoproject to the list of civic hacker organizations


jlord:
Awesome! Thank you @jungshadow!

Issue #54, opened 2013-10-16
community page should more a list style


muescha:
http://government.github.com/community/ should be more a list styel with the name of the organization additionaly - not only the logo

it gives a better overview

PS: for accessibilty is list also better

jlord:
@muescha I agree there should be another iteration on the /community page. We're learning there are so many more orgs and divisions than we imagined. 

I definitely want the page to useful and accessible so am going to let some more pull requests from orgs come in to get a better sense of the breadth of orgs and iterate on making the page better. 

For now, if you hover over the image, it will show the org name in the tooltip.

nschonni:
@jlord :+1: for the alt tag, but could it pull the GitHub `fullname` property instead?

benbalter:
>  for the alt tag, but could it pull the GitHub fullname property instead?

Right now, the only information we have is the org name, via `_config.yml`, and I'd love to keep it that simple, if possible.

> should be more a list styel with the name of the organization additionaly - not only the logo

We definitely can and should continue to improve the experience. For insight, part of the original vision was the effect of having a wall of logos to show just how large the community is.

nschonni:
> Right now, the only information we have is the org name, via _config.yml, and I'd love to keep it that simple, if possible.

KISS :smile:, not sure how hard it would be to use the GitHub API to get the info when _config.yml gets converted to the organization JSON file

muescha:
just now the github userpage dont give much information about the organisation

rmw:
I agree it'd be helpful to have more info about the organization - since they are adding into YAML anyway, we could just change the format to something like:

```
organizations:
  -
     github_name: awesomegov
     full_name: Super Awesome Gov
     location: Super Awesome Country
     short_description: blah
```

That way they can add if they want and we'll just show what's available.  It wouldn't be too much to convert the existing yaml to this either. OR we could check if the val is a hash or not and if not, only show the name.

Happy to take this on if people are cool with it.

nschonni:
@rmw I started something like that in #139, but feel free to take that base (or not) and improve upon it!

PR #240, opened 2014-01-19
Add the ability to see organizations in list form

rmw:
It was difficult to know what organizations were based only on their
avatar.  It was also difficult to understand the organization of the
organizations/countries in block format.  While imperfect, showing a
list with a github handle can give more information.
- Add toggle links for list/avatar view
- Add div with list to the community page
- Add id to the community avatar block
- Add styles for org list
- Add a small amount of javascript to show/hide the divs

### DIFF
--- a/assets/css/style.css
+++ b/assets/css/style.css
@@ -430,6 +430,18 @@ a.open.source:hover { color: #00acef; border-color: #00acef; }
   width: 80px;
 }
 
+#org-list ul {
+  list-style: none;
+}
+
+#org-list h2 {
+  border-bottom: 1px solid #999;
+}
+
+#org-list h3 {
+  border-bottom: 1px solid #EEE;
+}
+
 .avatar {
   height: 80px;
   width: 80px;
--- a/community.md
+++ b/community.md
@@ -6,7 +6,8 @@ permalink: /community/
 ---
 <div class="container">
   <div class="row">
-    <div class="span8">
+    <div class="span8" id="org-avatars">
+    <div><a href="#" id="view-list">View List</a></div>
     {% for type_hash in site.data.organizations %}
     <div class="type-block" id="{{ type_hash[0] | downcase | replace: ' ','_' }}"><p>{{ type_hash[0] }}</p></div>
       {% for org in type_hash[1] %}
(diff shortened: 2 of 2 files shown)


jlord:
It is hard to read with just the avatars, and while the avatars make for a nice open-gov-quilt, if someone wants to read them, a list would be more useful. Thanks for getting on this @rmw!

I made a [branch](https://github.com/github/government.github.com/tree/rmw-add-org-list) with your changes and made a few style edits. Here are screenshots so others can get a sense of the updates: 

_list view_
screen shot 2014-01-18 at 5 13 25 pm

---

_avatar view_
screen shot 2014-01-18 at 5 13 15 pm

This list does get really, really, really long and I think it could actually do for some real table formatting so people can sort and such. But I'm :+1: for this in the meantime.

_also_

I did try to have the avatars spill right so that over all the list was shorter and we could fill some of that white space, but because the names vary so much, it was pretty wonky:

screen shot 2014-01-18 at 5 19 37 pm

cc @benbalter

rmw:
Thank @jlord - looks great! I did my best to stay with the styles but happy to have another hand in on that.  Yea, it's unfortunate about the white space, but it's not horrible.  Similar reason I added the border/shadow (empty logos + right align looked weird.)

nschonni:
Removing the "github/" from the names would also thin down the grid

jlord:
@nschonni I added 'github/' because I think it helps identify them as user/org/account names, otherwise, it can look a bit alphabet soup because not all of the account names are very readable. Some of them, even without the 'github/' on the front are still longer than the images ('department-of-veteran-affairs'), so we'd still have the problem even without that prefix :/

nschonni:
For sure, unless you want to loop back to this idea https://github.com/github/government.github.com/issues/54#issuecomment-26468283

> it can look a bit alphabet soup 

Welcome to the government :wink:

benbalter:
>  I added 'github/' because I think it helps identify them as user/org/account names, 

Perhaps @? Would love to keep a grid-ish view for my OCD-ness. :smile:

jlord:
:+1: @benbalter I took out the 'github/'

Since some of the org names are still pretty long, not all of it aligns perfectly, but for the most part, it does the job:

_meh_
screen shot 2014-01-29 at 11 19 32 am

_better_
screen shot 2014-01-29 at 11 19 20 am

Overall though it's definitely much more readable.

benbalter:
@jlord are you still working in the [`rmw-add-org-list` branch](https://github.com/github/government.github.com/tree/rmw-add-org-list)?

jlord:
@benbalter do you mean is that where these updates are? Yep.

benbalter:
Oh man. This is great. Two questions:

> Add toggle links for list/avatar view

What's the value in the toggle? Right, if the original  problem was that "It was difficult to know what organizations were based only on their avatar", why don't we just :fire: the view that's difficult to use? (or is there a legitimate use case for both?) I'm just a big fan of not making options when a decision can suffice. It's just not super-intuitive to me because the initial load state almost looks like a list to me, and the post "view list" link (which has a "view avatars" toggle), contains avatars.

Second, this may break some unspoken CSS rule, but if the content area is 576px, can we just split that in four and have each row contain four equally spaced squares, content centered? (with some not having any content) I think that'd would allow everything to line up.

benbalter:
@rmw or @jlord any interest in picking this up?

jlord:
I made another branch the other day! https://github.com/github/government.github.com/tree/listview, it's very basic but I'll get a PR going :soon:

benbalter:
Closing in favor of #284, but far from finalized. Would still love your help making that page :sparkles:!

PR #241, opened 2014-01-27
Remove the reference to sitemap.xsl

haacked:
This xsl file doesn't exist so Chrome just renders a blank page when you visit sitemap.xml.


XhmikosR:
:+1:

benbalter:
Thanks @Haacked!

PR #242, opened 2014-01-28
Library updates

XhmikosR:


benbalter:
Believe https://github.com/github/government.github.com/pull/246 should implement this. Thanks!

PR #243, opened 2014-01-28
Added PROCERGS

guilhermednt:
PROCERGS is the data processing company for the brazilian state Rio Grande do Sul.

We are currently working on a few open source projects, one of which is already available on GitHub.


jlord:
Thanks @guilhermednt!

PR #244, opened 2014-01-28
Add Postcode to the fray!

waltz:


jlord:
I'd actually had a mental note to come on here and add y'all! Congrats it btw! Thanks @waltz!

waltz:
Thanks a bunch @jlord mergin' us! We're pretty pumped to be in the incubator!

On January 29, 2014 at 11:04:44 AM, Jessica Lord (notifications@github.com) wrote:

I'd actually had a mental note to come on here and add y'all! Congrats it btw! Thanks @waltz!

PR #246, opened 2014-01-31
Vendor third-party assets in Bower

benbalter:
For ultimate maintainability.

Even if slightly less performant, makes upkeep as simple as Bower update.


XhmikosR:
I don't want to sound like an ass, but seriously this was done very sloppy.

It wasn't properly thought or implemented. I don't see anyone having reviewed this either.
1. the bower components shouldn't be in the repo; just require `bower install`
2. jquery should be moved to bower.json and specify 1.11.0 otherwise v2.x doesn't support older IE
3. #229 is broken after this PR.

Generally this is a huge step backwards and I hate to see such changes go untested in a GitHub own repository.

benbalter:
> the bower components shouldn't be in the repo; just require bower install

What do you mean? The vendored assets need to be in the `gh-pages` branch for them to be served along side the site.

> jquery should be moved to bower.json and specify 1.11.0 otherwise v2.x doesn't support older IE

Good call. Will do.

> #229 is broken after this PR.

Is it just the jQuery issue?

XhmikosR:
Please don't push anything, I'm working on a fix as we speak.

XhmikosR:
@benbalter: I really wish this was reconsidered. It's messy, we end up with files we don't need, it's not just `bower update`.

Make a comparison of the repository before and after Bower and tell me honestly, if it's worth the trouble and mess. Even if you add a Ruby based solution to minify the files, the repository is still significantly messier.

See also the choosealicense.com PR. You used ZeroClipboard from Bower, the CSS files aren't used from bower, ZeroClipboard needs to load its own specific ZeroClipboard.swf version and so on.

Anyway, I cannot make anyone change their opinion, nor am I affiliated with GitHub myself, so I hope someone sees my point of view after all.

PR #251, opened 2014-02-11
Remove gsa-ocsit

nschonni:
The GitHub organization appears to have been removed


benbalter:
Great catch. Believe they combined with @gsa.

/cc @gbinal

PR #207, opened 2013-11-22
Work in progress: accessibility page

benbalter:
:warning: **WORK IN PROGRESS** :warning:

This pull request spikes out the initial structure of a potential `/accessibility` page describing GitHub.com, GitHub Enterprise, GitHub for Mac and GitHub for Window's compliance with §508 accessibility standards. The information in the pull request is to be considered an ongoing draft until merged if merged at all.

The format is adopted from the industry-standard [voluntary product accessibility template (VPAT)](http://www.state.gov/m/irm/impact/126343.htm). There's some more background on [the itic site](http://www.itic.org/public-policy/accessibility):

> The Voluntary Product Accessibility Template®, or VPAT®, is a tool used to document a product's conformance with the accessibility standards under Section 508 of the Rehabilitation Act. The purpose of the VPAT is to assist Federal contracting officials and other buyers in making preliminary assessments regarding the availability of commercial "Electronic and Information Technology" products and services with features that support accessibility. 

[Firefox](http://www.mozilla.org/en-US/firefox/vpat-2.html) has a great example of a VPAT written for Humans that we can emulate in part.

Much of the content still needs to be fleshed out, and it may make sense to loop in particular teams for some of the questions below.

PR #258, opened 2014-02-20
Accessibility page (VPAT)

benbalter:
This pull request creates an `/accessibility` page describing GitHub.com, GitHub Enterprise, GitHub for Mac and GitHub for Window's compliance with §508 accessibility standards. The information in the pull request is to be considered an ongoing draft until merged.

The format is adopted from the industry-standard [voluntary product accessibility template (VPAT)](http://www.state.gov/m/irm/impact/126343.htm). There's some more background on [the itic site](http://www.itic.org/public-policy/accessibility):

> The Voluntary Product Accessibility Template®, or VPAT®, is a tool used to document a product's conformance with the accessibility standards under Section 508 of the Rehabilitation Act. The purpose of the VPAT is to assist Federal contracting officials and other buyers in making preliminary assessments regarding the availability of commercial "Electronic and Information Technology" products and services with features that support accessibility. 

[Firefox](http://www.mozilla.org/en-US/firefox/vpat-2.html) has a great example of a VPAT written for Humans that we can emulate in part.
## Rendered Preview

(Click for full screen view)

vpat

(This pull request reopens https://github.com/github/government.github.com/pull/207 after a rogue merge.) 

/cc @github/accessibility


benbalter:
@nschonni I know you had weighed in a bit on #207, any feedback? Once this is merged, would love to learn more about and follow up with WCAG compliance documentation.

gbinal:
:+1: 

Great job with this.

devinreams:
:+1:

Terrific work!

jcarbaugh:
This is really great. Nicely done!

nschonni:
:+1: 
Sorry, since 508 is US only, I don't have any knowledge in that area.
I might suggest using an icon in addition to the colour to distinguish the supports values.

benbalter:
> I might suggest using an icon in addition to the colour to distinguish the supports values.

If ever there was a case for Emoji support on GitHub pages.... Great suggestion... will save that for a 0.2 when we can hopefully make it less US-specific.

benbalter:
I am going to go ahead and merge this, as it's hard to preview/discuss without something tangible to point to, but this is by no means set in stone, so if you have comments/feedback, please continue to comment as we continue to iterate on both the document and our compliance.

LaurentGoderre:
regarding the  "A method shall be provided that permits users to skip repetitive navigation links." we solve this in two ways. 

The first one is to have hidden links as the first elements of our pages that are available to keyboard users. When they are focused on they become visible. In our case we provide direct links to the main content as well as key navigation areas.

The second way we address this is by dividing the page in logical section (using html5 sectioning) and use hidden headings to describe the section's purpose. This allows users who use an adaptive technology that allows browsing by header to quickly access sections of the page.

The reason why we use these overlapping approach is that while the second approach is way more powerful, it depends on the user having adaptive technology which is not always the case, especially for sighted users who only use the keyboard.

Please visit this page for an example:
http://wet-boew.github.io/v4.0-ci/index-en.html

Try disabling stylesheet to unveil the hidden parts.

LaurentGoderre:
@nschonni they have the color and the text which is probably be the best combination.

nschonni:
@LaurentGoderre yeah the text probably is sufficient, but I was thinking colour blindness

LaurentGoderre:
This would fall in the usability side because the text describe the status better than the color, the color is just there to enhance the visual. Icons could make it better or worse (if it gets too crowded)

thomasgohard:
I'm not sure icons would help here as you have two levels of support for which there is no standard icon ('Not applicable' and 'Supports with exceptions').

As someone who is colour-blind, I can say that the colours for 'Supports' and 'Supports with exceptions' can be hard to distinguish in certain conditions. For example, depending on how far back the screen of my laptop is tilted.

That being said, since the text is there in full, it's not a problem. I'm still getting the information I need to get to use the page. If the text was not available in full, then it would be an issue.

benbalter:
Believe ths skip to content issue should be resolved. See https://github.com/github/government.github.com/pull/261.

LaurentGoderre:
Hi @benbalter it's a good start but there are a few things that makes this solution incomplete.

The first one is that the link is not the first element that you encounter on each page. On mnahy page where there are input fields, the input fields get the priority, then the rest of the content. This makes navigating with keyboard extremely confusing and would constitute a failure of WCAG 2.0 SC 2.4.3 Focus Order because elements are focus out of order and out of context.

The second issue is that the skip link is not visible when focused. This links is not only needed for screen reader users but also for sighted users who can't use a mouse. This item constitute a failure of WCAG 2.0 SC 2.4.7 Focus Visible. Also the off page positioning caused some issues in the past prompting us to change it to the following:

```
selector {
    height: 1px;
    width: 1px;
    clip: rect(1px, 1px, 1px, 1px);
    overflow: hidden;
    position: absolute;
    margin: 0;
}
```

I would also strongly suggest adding the hidden headers to allow heading based navigation, though I understand this would require some more intense engineering.

PR #259, opened 2014-02-20
Add Mexico and UdeG (Universidad de Guadalajara)

oscargalileo:
The UdeG is the Universidad de Guadalajara from Guadalajara, Jalisco, Mexico and use GitHub to colaborate in open source development.

### DIFF
--- a/_data/organizations.yml
+++ b/_data/organizations.yml
@@ -43,6 +43,9 @@ Germany:
   - gbv
   - lobid
 
+Mexico:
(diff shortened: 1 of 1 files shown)


benbalter:
@oscargalileo thanks for the pull request. So far, we haven't been including public universities, mainly because the challenges they face are not as closely related to those of e.g., government agencies, but you should check out http://education.github.com, and apply to be listed there!

PR #260, opened 2014-03-01
A quick summary of GDS blog post about git/github

jystewart:
A summary of the blog post I wrote for the GDS technology blog on how we use git/github and a link out to it.

### DIFF
--- a/_posts/2014-02-28-gds_source_control.md
+++ b/_posts/2014-02-28-gds_source_control.md
@@ -1,0 +1,24 @@
+---
+layout: story
+author: James Stewart, UK Government Digital Service
+title: How GDS uses Git and Github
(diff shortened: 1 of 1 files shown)


benbalter:
Wow. This is absolutely great. My only feedback would be to not assume that the reader know's what GDS is in the title, perhaps "Open collaboration at the UK's Government Digital Service"?

Other than that, just need to add an image (perhaps the logo?), which you could either do by adding to the image directory and updating the YAML front matter, or just drag/drop here and we can take care of getting it published.

Thanks for taking the time to write this up!

benbalter:
@jystewart Grabbed the gov.uk logo (let me know if there is a different one that you think is better), and posted, http://government.github.com/stories/gds-source-control/.

Thanks again for taking the time to put this together.

jystewart:
@benbalter that's great. thanks.

PR #261, opened 2014-03-08
Change skip to content to "supports"

benbalter:
`<a href="#start-of-content" class="accessibility-aid js-skip-to-content">Skip to content</a>` and `<div id="start-of-content" class="accessibility-aid"></div>` were added to GitHub.com (and soon GitHub Enterprise), + some CSS/javascript to hide the link from non-screen readers and and properly set focus on click.

Props @muan!

/cc @github/accessibility


LaurentGoderre:
@benbalter @muan See my comment https://github.com/github/government.github.com/pull/258#issuecomment-37187576

PR #263, opened 2014-03-20
added alv-ch to organizations

tdascoli:
We're the developer team for all web applications of the public employment service in switzerland at the State Secretariat for Economic Affairs SECO.


jlord:
Awesome!

PR #264, opened 2014-04-02
Add unioslo

tobiasvl:
Add the University of Oslo

### DIFF
--- a/_data/organizations.yml
+++ b/_data/organizations.yml
@@ -50,6 +50,7 @@ New Zealand:
 Norway:
   - difi
   - kartverket
+  - unioslo
 
 Puerto Rico:
   - commonwealth-of-puerto-rico


jlord:
Hi @tobiasvl!

Sorry if I'm missing it, but can you share the connection between the university and the government? This list is of governments or groups working directly for or with government, so I just want to double check!

tobiasvl:
Hey @jlord! Don't think you're missing anything, it might be I misunderstood a little. It's a public university (in Norway these are governmental), and the organization itself has a few repos with projects that are used in the educational sector. Feel free to reject if that's inappropriate. I work as a developer at the university and I'm a governmental employee, so I didn't consider it, but I guess other countries don't do it like the socialist Scandinavians!

afeld:
@tobiasvl Added the avatar to https://education.github.com instead – you may need to refresh a few times to see it, as they're randomized.  Thanks!

tobiasvl:
@afeld Thanks yourself!

PR #271, opened 2014-04-21
Update organizations.yml

KevinWhite:
Adding in lgss (shared service partnering with multiple uk authorities) with the code we have currently available.


benbalter:
Thanks! :tada:

PR #274, opened 2014-05-07
Split Governments and Civic hacker data

nschonni:
Should clarify where people need to edit. The civic hackers can now be split by country as well.

PR #275, opened 2014-05-21
Change /community to a list 

jlord:
This turns the community page into a list. We have so many now that the avatar wall isn't that helpful. I don't even think we need a 'switch to avatar wall view' option, I think we can just go with list from this point forward. This is a super simple Bootstrap list implementation of our data.

It might be nice to have the list better organized (other than just country/group) but this is the easiest to ship version. Maybe later we want to think about getting more info in the data. 

screen shot 2014-05-21 at 9 38 58 pm

### DIFF
--- a/assets/css/style.css
+++ b/assets/css/style.css
@@ -302,6 +302,12 @@ a.open.source:hover { color: #00acef; border-color: #00acef; }
 
 /* Organizations Section */
 
+.table td {vertical-align: middle;}
+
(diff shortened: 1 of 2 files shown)


nschonni:
Think your branch might be a little out of date. I split the data in #274

jlord:
Ah, indeed.

jlord:
Closing because I made a new branch!

PR #277, opened 2014-05-27
Fix 404 to NIST page

nschonni:
This looks like the same page, but @bruceravel might want to confirm


nschonni:
And Travis is now green again :wink:

bruceravel:
That is the correct link.
Cheers,
B

benbalter:
Thanks! :tada: :us:

PR #282, opened 2014-06-14
Add gocarta, Chattanooga, Tennessee's transit agency

junosuarez:


benbalter:
Thanks @jden perhaps it'd be better fit in cities? https://github.com/gocarta/government.github.com/blob/gh-pages/_data/governments.yml#L140

PR #284, opened 2014-06-18
Show Community in a Table

jlord:
Since there are so many and the wall of avatars has become too large (a good problem!). 

**This PR**
- puts the Govs and Civic Hackers into tables.
- adds basic search
  - I found something super simple in a JSFiddle actually and added some _no matches_ logic.

**Ponderings**
- The amount of avatars we have to go and fetch is a bit redonk now. It's not the worst, but it will only get larger.
- Would it be nice to have more information about each org? This will likely require some manual work but it may be worth it for a nice dataset to have.

**WIP**
I'm going to refine some of the styles a bit more, but wanted to get the convo started.

screen shot 2014-06-18 at 2 27 42 pm

screen shot 2014-06-18 at 2 27 57 pm

screen shot 2014-06-18 at 2 28 59 pm

screen shot 2014-06-18 at 2 28 17 pm

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -22,7 +22,6 @@
       </div>
     </footer>
 
-  <script src="/assets/js/jquery.min.js"></script>
   <script src="/assets/js/bootstrap.min.js"></script>
   </body>
 </html>
--- a/_includes/header.html
+++ b/_includes/header.html
@@ -9,6 +9,7 @@
   <link rel="shortcut icon" href="/assets/img/favicon.ico" type="image/x-icon">
   <link rel="stylesheet" href="/assets/css/style.css">
   <link rel="stylesheet" href="/assets/css/icons.css">
+  <script src="/assets/js/jquery.min.js"></script>
   <!--[if lt IE 9]>
     <script src="/assets/js/html5shiv.js"></script>
     <script src="/assets/js/respond.min.js"></script>
--- a/_layouts/support-page.html
+++ b/_layouts/support-page.html
@@ -44,17 +44,21 @@
         $('.searchable tr').filter(function () {
             return rex.test($(this).text());
         }).show();
+        console.log("gov", $('.govtable tr:visible').length)
+        console.log("civ", $('.civictable tr:visible').length)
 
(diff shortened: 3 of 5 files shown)


benbalter:
Really digging this. 

Is it possible to sneak an anchor into the first org of each "affiliation". I deep link a fair amount to, e.g., `/community/#u.s._cities`

As for the pondering, I'd suggest opening an issue for each... let's get this merged and iterate... already a huge improvement.

benbalter:
One thought, if it's easier, we could sneak the array of orgs into the page as javascript (rather than HTML) and push everything client side for initial load, if it'd simplify the logic a bit?

jlord:
:thumbsup: adding more anchors

Confused by the second comment though, which logic are you referring to?

benbalter:
> which logic are you referring to?

Just kidding. Misread. Carry on. :smile:

jlord:
lol

Another option, we could paginate tables, this page does get looooooong (24145 px). But we can do that as another iteration/PR.

jlord:
Ok, to anchor to the first of each country I need to add a new row in. But then it seemed silly to have the name of the country at the top and have it repeated every row after on the right -- but there are some country lists that are long and you could potentially scroll and not know which country you were in. So I've kept the right side country name but lightened it so it's not so obviously redundant. But I'm still iffy on it. 

screen shot 2014-06-19 at 11 56 48 am

Could also just not put any text in the type row and make it look more like it did before:

screen shot 2014-06-19 at 12 04 52 pm

Hrmm.

benbalter:
I like 1 > 2, although we could get rid of the right column at that point. 

Why not just sneak an `<a id="FOO">` in with the first cell of each group?

jlord:
Alrighty! Now we've got:
- section blocks with id's you can link to! `/community/#u.s._city`
- cities are now blue cause I realized they should be links 

screen shot 2014-06-19 at 3 38 54 pm

benbalter:
:sparkling_heart: Such an improvement. If you're happy with it, I say :shipit:!

jlord:
Ready, set, :ship: !

PR #286, opened 2014-07-02
Update governments.yml

aepstein:
LANL is a Los Alamos National Lab, United States organization, not Australia.


benbalter:
Great catch! Thanks!

PR #289, opened 2014-07-02
Update governments.yml

aepstein:
Tasmania is a state of Australia.  Since other Australian state and local agencies are listed under Australia, so should Tasmania municipalities.


benbalter:
the more you know

PR #291, opened 2014-07-08
Update governments.yml

aepstein:
portofsandiego is a special district in California, not a federal agency.


benbalter:
Great catch! Thanks again @aepstein!

PR #294, opened 2014-07-29
Link to Government Showcase, fixes #293

### LINKED ISSUE #293: Link out to government showcase
Via @adamjacobbecker, we should link out to [the government app showcase](https://github.com/showcases/government) from government.github.com in a big way.

Mirroring the features "learn more" link, thinking it'd make the most sense to add it below the "how to use GitHub" section. Don't :heart: the copy, but something like:

screen shot 2014-07-29 at 9 50 31 am

@jlord, any suggestions?

jlord:
This adds a link to the Government showcases re #293 

What do you think about the copy? Should 'App' be used at all? Should 'Government __ Showcase' be lowercase?

screen shot 2014-07-29 at 12 45 15 pm

### DIFF
--- a/index.html
+++ b/index.html
@@ -83,6 +83,7 @@ org_count: 60
     </div>
 
   </div>
+  <p><a href="https://github.com/showcases/government">See more in the Government App Showcase &rarr;</a></p>
 </div>
 
 
--- a/index.html
+++ b/index.html
@@ -83,7 +83,7 @@ org_count: 60
     </div>
 
   </div>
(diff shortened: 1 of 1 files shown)


benbalter:
I like it. Only other idea is just "See more in the government showcase"? (to give us flexibility to showcase non-app things)

jlord:
Agreed :+1:

benbalter:
:sparkling_heart: :shipit:!

PR #297, opened 2014-08-01
Add count to org list

benbalter:
This is helpful to say, "oh, there are N US Agencies are on GitHub", without needing to count.

Looks a big ugly, but perhaps right aligned, and grey? I dunno. @jlord what do you think?

screen shot 2014-08-01 at 6 34 27 pm

### DIFF
--- a/assets/css/style.css
+++ b/assets/css/style.css
@@ -316,14 +316,11 @@ li {
 
 .search-section { padding-bottom: 20px; }
 
-/*.boxed {
-  background: #F8F8F8;
-  padding: 20px;
+.count {
+  color: #A3A3A3;
+  font-weight: 200;
+  font-size: 14px;
 }
-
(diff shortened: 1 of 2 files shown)


jlord:
I think adding a count is a great idea!

Playing with the style some, but I'm not sure if this is just crazy:

screen shot 2014-08-03 at 8 02 57 pm

benbalter:
I like it. Perhaps " - N organizations" or " - N orgs"?

jlord:
@benbalter :+1: yep! Just pushed commits:

screen shot 2014-08-04 at 12 46 47 pm

benbalter:
CI is failing due to the `.ruby-version` file. Will take a look. Otherwise, :+1:. 

Thanks, once again, for de-uglifying my code! :smile:

PR #299, opened 2014-08-06
Prune featured orgs to those with at least one public repo

randomecho:
Quick scan of the featured and the deleted organisations don't have any public repos, so landing on them as an example feels like a dead end for now. Leaving the rest to those with at least a repo showing under their org.


benbalter:
Great call @randomecho thanks!

PR #305, opened 2014-08-09
Adding @GovBrasil account name to the list

valessiobrito:


benbalter:
Tests were failing because the featured org image wasn't in the repo, but I removed it from the featured org list, and merged. Thanks! :smile:

PR #310, opened 2014-08-22
Add datagovhk

benbalter:
This pull request adds @datagovhk to the list of government organizations. They're doing some great open source work around data.gov.hk.

Going off of [the wikipedia entry](http://en.wikipedia.org/wiki/Hong_Kong), should Hong Kong be listed as it's own top-level category, or is China more appropriate?

(Ignore the @kentcc noise that came from running the alphabetization script before committing)

### DIFF
--- a/_data/governments.yml
+++ b/_data/governments.yml
@@ -102,6 +102,9 @@ Germany:
   - gbv
   - lobid
 
+Hong Kong:
+  - datagovhk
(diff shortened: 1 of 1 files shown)


benbalter:
Gonna merge as is, but glad to revisit if anyone feels strongly either way.

PR #311, opened 2014-08-22
Remove usgin-modelsw

nschonni:
Appears to be a dead org and is causing the Travis build to fail


benbalter:
Huh. Could have sworn I did that. Must have clobbered it. Thanks as always @nschonni!

nschonni:
I think the alphabatizing accidentally confused the diff and added it back

PR #312, opened 2014-08-22
adding @codefordc

leahbannon:


benbalter:
Thanks @leahbannon! :tada: :us:

PR #313, opened 2014-08-22
Adding 12 more brigades

leahbannon:
All brigades listed as official brigades in http://www.codeforamerica.org/brigade


leahbannon:
make that 15

jlord:
Oh, awesome! Thanks so much @leahbannon! :sparkles:

PR #314, opened 2014-08-27
adding codeforseoul

hoony-o-1:
- Code for Seoul is a civic hacking group in South Korea


jlord:
Awesome, thanks @thechunsik!

PR #319, opened 2014-09-20
Explain in abbreviation tag who GSA is

randomecho:
Guessing GSA means the "General Services Administration" in this context


jlord:
Thanks! @randomecho!

PR #321, opened 2014-09-24
Add the new one pager to site

jlord:
This adds the one pager we made to the site:

screen shot 2014-09-24 at 3 00 23 pm

cc @benbalter


nschonni:
:sparkles: : I think it looks good, but why not code it in HTML?
:trollface: Is a PDF really an `img` asset

benbalter:
@nschonni the idea is that the one-pager is essentially a PDF version of the landing page. One of the most frequent requests we get (especially around open source code) is for resources that developers can (print and) give to their bosses to explain what GitHub, open source, etc. is.

nschonni:
> is for resources that developers can print

Maybe tackle it as a print stylesheet then :wink: You're right that the frontpage looks really rough in print preview (but great on screen).

jlord:
>  print stylesheet then

It is a lot like the front page, but we didn't want to take the front page exactly and it has some extra bits too that don't exist on the front page. 

Also, we needed to create it for a booth we had at a conference so now that it already exists, why not drop it on here. 

As it's fairly harmless, and we can always do something else later, gonna merge.

PR #322, opened 2014-10-01
Update governments.yml

grothja:
Requesting that NIAID be added to the government.github.com site.


benbalter:
Thanks @grothja! :tada:

PR #327, opened 2014-11-07
Add ApieFrance to governments.yml

pimpin:
Apie stands for "Agence du Patrimoine Immateriel de l'Etat". It means _Agency for public intangibles of France_


benbalter:
Thanks @pimpin!

pimpin:
Thanks you @benbalter.

PR #331, opened 2014-12-11
Update governments.yml

ghost:
Added organization Comune di Bologna

www.comune.bologna.it

For info please refer to:

admin@comune.bologna.it


benbalter:
@ComuneBologna it looks like @ComuneBologna is a user, not an organization. I believe you'd want the City to have an [organization account](https://help.github.com/articles/what-s-the-difference-between-user-and-organization-accounts/), not a user account, before being listed.

PR #332, opened 2014-12-12
Update governments.yml

gamici:
Added Comune di Bologna
www.comune.bologna.it
refer to: opensource@comune.bologna.it


jlord:
Awesome!

PR #334, opened 2014-12-18
added NYCEDC (http://edc.nyc) to the list

mralexho:
NYCEDC has several datasets produced on a monthly basis from the Research & Analysis team. Datasets include Labor Force & Payroll, Travel & Tourism, Demographics, Real Estate & Construction, and more.

We'll be releasing these datasets and tools in the coming months.

### DIFF
--- a/_data/governments.yml
+++ b/_data/governments.yml
@@ -221,6 +221,7 @@ U.S. City:
   - cityoffortworth
   - CityOfGreeleyCo
   - cityofnewyork
+  - NYCEDC
   - CityOfPhiladelphia
   - CityofReno
   - CityofSanFrancisco


benbalter:
@mralexho is @NYCEDC a government agency? A non-profit?

mralexho:
@benbalter it's a hybrid setup but for simplicity sake, it should be non-profit. I'm not entirely sure where I can place @NYCEDC. Would we fit under civic hackers?

benbalter:
Civic hackers would probably make more sense, if you're able to make the change.

nschonni:
@jbjonesjr this already landed in #335

PR #337, opened 2015-01-21
Update governments.yml

ivbeg:
Moved GovBrasil from Bolivia to Brasil list of repositories


benbalter:
Believe that was my fault. Thanks @ivbeg! :smile:

PR #338, opened 2015-01-22
Added @gsi-cyberjapan

johofukyu:
Added @gsi-cyberjapan, Information Access Division of Geospatial Information Authority of Japan.
We are a division of the national geospatial information authority, or national mapping agency, of Japan.
See https://github.com/gsi-cyberjapan/gsimaps for an example of our activities.

### DIFF
--- a/_data/governments.yml
+++ b/_data/governments.yml
@@ -132,6 +132,9 @@ Hong Kong:
 India:
   - itschool
 
+Japan:
+  - gsi-cyberjapan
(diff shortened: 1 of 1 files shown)


benbalter:
Hard to believe this is the first Japanese organization on the list. Thanks @johofukyu! :japan: :jp:

johofukyu:
Thanks @benbalter !

PR #339, opened 2015-01-23
adding codeforibaraki

tosato3:
Code for Ibaraki is a civic hacking group in Japan.


benbalter:
Thanks @tosato3! :jp:

PR #343, opened 2015-02-09
Update governments.yml

matthew-shaw:
Removed internal Land Registry orgs


benbalter:
Thanks @matthew-shaw!

PR #345, opened 2015-02-17
Add @SGMAP-AGD to list of French GitHub users

MattiSG:
Sorry, I learned about that one only today  ;)


benbalter:
Thanks @MattiSG! :fr:

PR #346, opened 2015-02-20
adding wakayama-pref-org

sakano-y0003:
@wakayama-pref-org is a government agency in Japan.


benbalter:
Thanks @sakano-y0003! :japan:

PR #347, opened 2015-02-26
Update governments.yml

japurcell:
Requesting to add my organization to the list.


benbalter:
Thanks @japurcell. Googling, it looks like you're "Puget Sound Naval Shipyard & Intermediate Maintenance Facility"? Is that correct? If so, perhaps it'd make more sense in the next section, US military? Also, it might be nice to add the name to your organization's profile (and logo) if you can!

PR #353, opened 2015-04-02
Update governments.yml

kelpisland:
Added BCGov (Province of British Columbia, Canada). We're excited to be here :)

Todd


benbalter:
Thanks @kelpisland! :maple_leaf:

PR #354, opened 2015-04-02
Update governments.yml

barthanssens:
Added Fedict (Federal Public Service for ICT in Belgium)


benbalter:
Thanks @barthanssens!

PR #358, opened 2015-04-10
Deleted swiss-gov from Colombia

montogeek:


benbalter:
Thanks @montogeek. I believe that was my mistake. Nice catch.

montogeek:
My pleasure!

PR #361, opened 2015-04-15
Move Brighton and Hove to UK Councils

CraigGardener:
Brighton and Hove is a "UK Council" not a "UK Central Government" organisation.


benbalter:
Thanks @CraigGardener! :gb:

PR #376, opened 2015-06-19
Update governments.yml

sniyozov:


sniyozov:
Add new gov repository (Tajikistan)

jbjonesjr:
@sniyozov can you combine this with the efforts in #379 as well?

PR #378, opened 2015-06-19
Update governments.yml

sniyozov:
Add OpenDataTajikistan to government repository list

### DIFF
--- a/_data/governments.yml
+++ b/_data/governments.yml
@@ -254,6 +254,9 @@ Switzerland:
   - sitn
   - swiss-gov
   - vdch
+  
+Tajikistan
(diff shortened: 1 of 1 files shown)


_data/governments.yml
```diff
@@ -254,6 +254,9 @@ Switzerland:
   - sitn
   - swiss-gov
   - vdch
+  
+Tajikistan
```
benbalter, line 258: It looks like you're missing a `:` after `Tajikistan`

nschonni:
Dupe of #379 ~~which has the fixed format~~ (Edit: nevermind, it's the same)

PR #383, opened 2015-07-01
Update civic_hackers.yml

sniyozov:
Add new civil hackers group from Tajikistan


benbalter:
Thanks @sniyozov! :tada:

PR #385, opened 2015-07-14
Just use site.title on the home page

chibicode:
Before, when viewing the home page

``` html
<title>GitHub and Government | GitHub and Government</title>
```

After:

``` html
<title>GitHub and Government</title>
```


benbalter:
Nice catch. Thanks @chibicode!

chibicode:
@benbalter Thanks!

PR #386, opened 2015-07-15
Add miambiente to governments.yml

rcano-miambiente:
Add **Ministerio de Ambiente de Panamá** to the list of government institutions.


benbalter:
Thanks @rcano-miambiente!

PR #392, opened 2015-09-04
Fix AusDTO

maxious:
AusDTO organisation was renamed.


benbalter:
Thanks @maxious! :cactus:

PR #400, opened 2015-09-22
Fix travis build

sonalkr132:
Probably, using different names would be a better idea?


benbalter:
Ahh nice catch. Thanks @sonalkr132. :link: :white_check_mark:

PR #404, opened 2015-09-23
Use images of correct size from github api

sonalkr132:
I think this should improve the speed of index page as now avatars are originally 80px wide (73px + 6px padding). 

Currently, [government.github.com](https://government.github.com) is using images from repo which are 460px wide, inline width property is resizing them after loading the full size image.

**UPDATE:**
Some relevant info from chrome debugger (no cache):
[site using 80px avatars from api](http://sonalkr132.github.io/government.github.com/): data transferred - 1.1 MB and Finish time - 5.78 s
[current gov site](https://government.github.com/): data transferred - 6.6 MB and Finish time - 14.72 s


sonalkr132:
**Improvement info about community page** (no cache):
_sonalkr132/img-size_ branch: requests made - 752, data transferred - 2.4MB,   Finish time - 17.22s
_government.github.com/gh-pages_: requests made - 1490, data transferred - 38.8MB,   Finish time - 1.3m

_government.github.com/gh-pages_ is currently using `https://github.com/{{ org }}.png` which gets redirected to:

```
"https://avatars0|1|2|3.githubusercontent.com/{{ org }}?v=3
```

thus ends up making two requests for each org icon.

17.22s is still a bad time but at least it's an improvement.

**EDIT**
correct url

benbalter:
@github/storage is this a safe way to request a specific size of an org's avatar by URL, knowing is only its username? 

We'd like to be able to use vanilla HTML (via Jekyll) to list all government orgs, but right now we're using `github.com/username.png` which can't be sized. Wasn't sure if the `assets3` URL was stable or may disappear in the future.

technoweenie:
https://avatars.githubusercontent.com/technoweenie?s=10

You can also use avatars0-avatars3 for the subdomain. This is a trick to get current browsers to load a bunch of images faster. Since this is a static site, I don't think it's necessary (or even possible).

_includes/org-table.html
```diff
@@ -16,7 +16,7 @@ <h6 class="no-matches" style="display: none;">No matches.</h6>
         <tr>
           <td>
             <a href="https://github.com/{{ org }}" title="{{ org }}">
-            <img src="https://github.com/{{ org }}.png" width="40" height="40" alt="{{ org }}"></a>
+            <img src="https://avatars.githubusercontent.com/{{ org }}?v=3&amp;s=40" width="40" height="40" alt="{{ org }}"></a>
```
technoweenie, line 19: NVM! I must've been seeing things :)

<del>This should just be `avatars.githubusercontent.com`. It'll work, but you want to keep the host names consistent across your site so browsers and CDNs aren't storing multiple copies in their caches.</del>


<del>On a dynamic site, I'd suggest using consistent hashing on the path (`/org`) to get a consistent modulus for the subdomain (`avatars0`-`avatars3`). I don't think we can do that here without some kind of custom plugin, so I'd recommend just using `avatars`.</del>

benbalter:
Nice. Thanks @technoweenie and thanks @sonalkr132!

_includes/org-table.html
benbalter, line 19: Let me play around and see if I can figure something out. May be able to do a mod on the org name length or something.

sonalkr132, line 19: @technoweenie You weren't seeing thing :stuck_out_tongue: 

I rebased it.

PR #406, opened 2015-09-25
use avatars0-3 based on org name length

benbalter:
A follow up to https://github.com/github/government.github.com/pull/404#discussion_r40442384, this uses the `avatars0` through `avatars3` subdomains for loading avatars by calculating mod 4 of the length of the org name. A bit of a hack, but it works to consistently generate 0-3 in a somewhat random distribution.

/cc @technoweenie and @sonalkr132!

### DIFF
--- a/_includes/org-table.html
+++ b/_includes/org-table.html
@@ -16,7 +16,9 @@
         <tr>
           <td>
             <a href="https://github.com/{{ org }}" title="{{ org }}">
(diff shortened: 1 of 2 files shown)


technoweenie:
Noooice :+1: You can prove this works by making sure the avatar subdomains stay the same on multiple page rebuilds.

FWIW This is what we use to get this on GitHub:

``` ruby
suffix = "/user?s=123"
Zlib.crc32(suffix) % 4
```

sonalkr132:
LGTM!

PR #411, opened 2015-10-12
Update civic_hackers.yml

sommanita:
We are a Code for America brigade and would like to enable private repos. We are Code for Greensboro (http://github.com/codeforgso).


benbalter:
Thanks @sommanita!

PR #422, opened 2015-12-09
Added skat-repositories to list

JaneDahl:
Please append SKAT (Danish Tax and Customs Administration, http://www.skat.dk) to the list of government organizations.


benbalter:
Thanks @JaneDahl!

PR #423, opened 2015-12-09
Downgrade HTML proofer

benbalter:
See the output of https://travis-ci.org/github/government.github.com/builds/95851068.

It seems HTML Proofer `2.6.0` breaks the build. I suspect it's related to the escaping in https://github.com/gjtorikian/html-proofer/pull/273.

/cc @gjtorikian


gjtorikian:
Yep, that's the PR. Damn. Shouldn't have tried to shortcut it with regexps. Fix incoming, I hope.

PR #425, opened 2015-12-11
add the City of Nanterre

PLrolle:
adding the City of Nanterre's github organization page to the french local government list.


benbalter:
Thanks @PLrolle!

PR #438, opened 2016-02-26
Community page refresh

benbalter:
This pull request is a refresh of the `/community/` page, to make it a 2D grid of avatars, rather than a long vertical table. The idea being, is that as the list grows (:tada:) it should theoretically be a better use of screen real estate, and in my mind, a big grid of avatars is more impressive than a long table, since you can see more at any given time.
### Before

screen_shot_2016-02-26_at_11_33_54_am
### After

screen shot 2016-02-26 at 11 30 54 am

and

screen shot 2016-02-26 at 11 30 59 am
### A few things to call out
- Javascript was converted to CoffeeScript, and is only loaded on the Community page
- Javascript moved to the footer
- Search works just as before
- All existing anchors should continue to work

Fixes https://github.com/github/government.github.com/issues/54.


XhmikosR:
@benbalter: you don't like rebasing at all :P

Seriously though, it makes a lot more sense to not have all those merges when you are working on a dev branch.

XhmikosR:
Is it really worth adding coffeescript just for those 2 files? The rest totally make sense.

PR #444, opened 2016-02-26
Minor CSS cleanup.

XhmikosR:
- remove unused selector
- use a shorthand
- fix `background-size` for `.full-width.olives`

@benbalter: I went with `190%` for `background-size` [here](https://github.com/github/government.github.com/commit/78be3ddc1290217e2bacb600017560a5d34db11b#diff-72abf624385ed0401c0c530691c3ac3bR447). If something else was meant, let me know.

### DIFF
--- a/assets/css/style.scss
+++ b/assets/css/style.scss
@@ -16,7 +16,7 @@ li img {
 /* begin overriding */
 
 body {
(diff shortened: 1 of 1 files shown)


benbalter:
Tested locally and looks good. Thanks, as always, @XhmikosR!

PR #447, opened 2016-02-29
_config.yml: exclude more unneeded files.

XhmikosR:


benbalter:
`organizations.json` is actually a Jekyll-generate file, used to expose the organization list in a machine-readable way.

XhmikosR:
Fixed.

benbalter:
Thanks!

PR #449, opened 2016-02-29
Use `https` URLs when possible.

XhmikosR:
Also, update docs for GitHub desktop and remove unneeded html-proofer suppressions.

If tests pass, it should be good to go @benbalter


benbalter:
Looks good. Thanks @XhmikosR!

PR #459, opened 2016-04-29
Update governments.yml

hannesj:
Add Ruter AS (Public transport authority of Oslo) and Rutebanken AS (National authority for coordination of public transport information in Norway, owned by vegvesen)


benbalter:
Thanks @hannesj!

PR #460, opened 2016-04-29
Add two Finnish organizations to organizations.yml

jmerikko:


nschonni:
@jbjonesjr this should be web rebaseable

PR #466, opened 2016-05-20
Add mxabierto to list

EduardoClark:
mxabierto is the repository for the Mexican open data policy


benbalter:
Thanks @EduardoClark!

PR #470, opened 2016-06-01
add ECCC-MSC

tomkralidis:
Add Meteorological Service of Canada.


benbalter:
Thanks @tomkralidis! ⛅

PR #476, opened 2016-07-01
add two more agencies

afeld:


benbalter:
Nice. Thanks @afeld! 💪 💻 :us:

PR #479, opened 2016-08-05
Update research.yml

jpiburn:
Adding GIST-ORNL to the list

### DIFF
--- a/_data/research.yml
+++ b/_data/research.yml
@@ -15,6 +15,7 @@ U.S. Research Labs:
   - EMSL-MSC
   - epicsdeb
   - exmatex
+  - GIST-ORNL
   - GENIEMC
   - globus
   - HipGISAXS


jbjonesjr:
👋 @jpiburn . Not to question you, but is this a research group or a govt org? 

I guess it's some sort of Hybrid? (I'd classify ORNL as a govt org, but i guess individual projects under it are research orgs?). Any thoughts?

jpiburn:
@jbjonesjr ORNL is part of the Department of Energy laboratory system and as such we are classified as a federally funded research and development center. We work with other government agencies as well as industry providing basic and applied research in energy and security. So to answer your question we are a government organization focused on R&D

jpiburn:
@jbjonesjr sorry I suppose I forgot to address what the GIST-ORNL group is. The GIST group is a group within the organizational structure of the lab focused on geographic information science. Everything that applies to ORNL from the above comment applies to the GIST group as well. We are a government organization focused on R&D

jbjonesjr:
@jpiburn ok, thanks for the details. in that case, i'd argue to move this commit from the research file to the `governments.yml` file. Seem reasonable?

jpiburn:
@jbjonesjr sounds like plan. Thanks!

PR #482, opened 2016-08-24
Added igov.org.ua organization

kilotaras:
Non Commercial social-governmental startup for fighting with corruption in Ukraine.

Tries automatize as much burecracy as possible.


benbalter:
Thanks @kilotaras!

PR #487, opened 2016-09-20
Added ICANE's profile to organizations list

predicador37:
Added Instituto Cántabro de Estadística's profile, the statistical office of Cantabria (a region from Spain) belonging to the Government of Cantabria.


predicador37:
Travis is failing due to not found external URLs that have nothing to do with the organization I've added. Could anyone take a look?
Thanks in advance!

PR #488, opened 2016-09-23
Update to governments.yml

bearpig:
Added Jersey and the statesofjersey account.


bearpig:
This build failing seems to be occurring with a lot of other Pull Requests. Can you check it's not an issue that's not to do with adding new entities to the governments.yml, or something wrong with the 'add entity' readme?

PR #493, opened 2016-10-08
Added several new repositories

ivbeg:
Added repositories for several official government opendata projects in Austria, Croatia and Luxemburg


nschonni:
Looks like Travis is failing because `datagvat` and `datagovhr` are users, instead of orgs

ivbeg:
Probably we could add not only organizations and to fix travis ?

PR #496, opened 2016-10-14
Update civic_hackers.yml

ivbeg:
Added multiple organizations related to Open Knowledge Int. (especially to OKF DE) and other major civic open data related projects.


jbjonesjr:
Awesome additions @ivbeg . Thanks for the assist!

PR #499, opened 2016-10-24
Add Commerce Western Australia to the list.

ocean:
Online Services at the Department of Commerce Western Australia has some projects on GitHub now, so I've added us to the list.


benbalter:
Thanks!

PR #500, opened 2016-10-24
Fixed broken link to Singapore Gov Tech

ocean:
The username for this org has changed so the CI link check was breaking. Fixed it now.


benbalter:
Thanks again @ocean!

PR #501, opened 2016-10-24
Removed Data Seattle.

ocean:
Seattle Data Gov org has disappeared, so the CI link checker was breaking. Removed the link.


benbalter:
Thanks @ocean!

PR #503, opened 2016-10-27
Update civic_hackers.yml

josephbleroy:
Added https://github.com/openopps to Civic Hackers.

> OpenOpps is the open source platform that supports the Open Opportunities program. As open source software, it can be deployed by state or local governments or in the private sector. It is free to use by anyone who can has technical operations staff and passion to start a new program, and we welcome everyone who would like to contribute to improving the platform.


nschonni:
Since it looks like this is done by 18F, wouldn't it be under US Federal?

josephbleroy:
@nschonni I wasn't sure. I had the same thoughts, but it seemed like a completely separate entity. I know 18F would be the organization, but OpenOpps seems like it's the open source group that built the platform?

nschonni:
@rogeruiz since you're the only person visible in the org, maybe you can answer where it should go

rogeruiz:
Ah yes! First, thank you for pinging me. Second, let's see how much I can answer based on the above and my experience on the team.

Open Opportunities Platform was once owned and operated by 18F and that was my first experience with the project. It has since moved to an open source governance model outlined at the root of the repository.

The Open Opportunities program, which is the deployed federal version of the platform is still owned and operated by a government agency but no longer by 18F. The code is open source and now lives in its own Github organization. 

I think adding it under Civic Hackers is a good call. The move to an open source governance model was an effort by former 18F colleagues to ensure that playform could be actively worked on outside of 18F. There hasn't been much engagement within the core team due to time commitments with other things, but I leading an effort to get started doing more for the platform in my non-government capacity.

@nschonni

jbjonesjr:
Thanks @rogeruiz for sharing those details. I think this makes sense as a civic hacking platform as you mentioned, rather than a govt org. Again, this is an example where I wish we could provide a bit more details about these orgs and what they do... maybe the next revision of the site.

I'm going to go ahead and update the branch @josephleroy and merge this pr. Thanks for the contributions.

PR #507, opened 2016-11-01
Add 2 London councils

BforBen:
Add Royal Borough of Kingston-upon-Thames and London Borough of Sutton


jbjonesjr:
thanks @BforBen . I've been merging a ton of UK govt and locality orgs the past few days. Do you guys have a group where you coordinate together?

PR #509, opened 2016-12-01
Update GCHQ organisation name

jystewart:
GCHQ are now releasing code as "GCHQ" rather than "GovernmentCommunicationsHeadquarters"


nschonni:
I'm looking at the failure in #510. Looks like a bundler thing, and now some dead orgs

PR #513, opened 2016-12-11
Consolidated US Federal and US Military and Intelligence in governments.yml

### LINKED ISSUE #463: Consolidate U.S. Federal & U.S. Military and Intelligence in governments.yml
No other country breaks their list of "Federal" accounts this way and don't see what added value there is as source code and projects can find re-use across military and civilian sectors. This only adds un-needed complexity to keep the list up to date for the US having to compare three lists in what should be two list within two files:
- https://github.com/github/government.github.com/blob/gh-pages/_data/governments.yml
- https://github.com/github/government.github.com/blob/gh-pages/_data/research.yml

Related:
https://github.com/github/government.github.com/pull/461

ryancharris:
Closes #463

### DIFF
--- a/_data/governments.yml
+++ b/_data/governments.yml
@@ -505,6 +505,7 @@ U.S. Federal:
   - adl-aicc
   - adlnet
   - afrl
+  - afseo
   - arcticlcc
   - BBGInnovate
   - bfelob
@@ -524,6 +525,7 @@ U.S. Federal:
   - EEOC
   - energyapps
   - erdc-cm
+  - erdc-itl
   - eregs
   - fcc
   - FCCdev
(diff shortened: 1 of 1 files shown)


benbalter:
The rationale for breaking them up was two fold:

1. The US federal list is already the largest, and we were looking to break it up into smaller subsets to keep it more manageable.

2. The benefit of breaking it into a distinct list was for deep-linkability, when answering the question "are there any military or intelligence agencies using GitHub"?

I don't feel strongly, if there's a strong motivation to collapse them into one other than data consistency.

jbjonesjr:
@ryancharris any thoughts on why you want to merge the two? I agree with @benbalter's status, but am interested in your thoughts

ryancharris:
@jbjonesjr After reading @benbalter's feedback, I actually agree with him that leaving the two as separate lists may be the best way to go, especially given the fact that the U.S. Federal list is the longest.

While the suggestion in #463 to reduce complexity seemed like a logical move at first, it may actually have the opposite effect by making the combined list somewhat unmanageable.

jbjonesjr:
:+1: thanks @ryancharris. Feel free to reopen if you want later.

miaguila0710:
My soon need the support to UUEE Because In Spain u image is denigrant to the government statement

PR #516, opened 2016-12-16
Added entry for UKGovernmentBEIS

DougC:
Was pointed in the direction of this list by Anna Shipman at GDS


jbjonesjr:
thanks @DougC , and 👋 @annashipman , thanks for the pointer!

PR #517, opened 2016-12-17
Add GovITHub to the list 

ClaudiuCeia:
http://ithub.gov.ro/


ClaudiuCeia:
In governments.yml, in the France group, the following entries are users, not orgs:
GouvernementFR
In governments.yml, in the U.S. Federal group, the following entries are users, not orgs:
USDA-NRCS

Seems like an unrelated issue, #518 fixes it.

PR #520, opened 2016-12-20
Update governments.yml

OutKa5t:
Added Skills Funding Agency https://github.com/SkillsFundingAgency


_data/governments.yml
```diff
@@ -399,6 +399,7 @@ U.K. Central:
   - openglasgow
   - osgeouk
   - scottishgovernment
+  - Skills Funding Agency
```
nschonni: There are no spaces in the org name

jbjonesjr:
Thanks @OutKa5t . What a lively org that is! Would love to see some pinned repos or maybe a github pages site for high level organization, but a great example for other UK organisations. Thanks again for sharing, and let me know if there is anything I can do to help.

PR #528, opened 2017-01-16
Cleanup orgs for the build to pass

nschonni:
Org appears to be deleted and is failing the build now

PR #530, opened 2017-01-17
Run the Alphabetize script

nschonni:
Automated reordering of the Yaml data with the `scripts/alphabetize`

### DIFF
--- a/_data/civic_hackers.yml
+++ b/_data/civic_hackers.yml
@@ -1,125 +1,125 @@
 # Add non-profits or groups focused on government
 
 Civic Hackers:
+  - 16iacc
+  - adhocteam
   - akroncivichackathon
+  - avoindata
+  - BD-IATI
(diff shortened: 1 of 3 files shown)


jbjonesjr:
ugh, i'm so torn, because i love things being in alphabetical order, but also love grouping states and similar orgs together (why doesn't everyone prefix their `org names` with their higher level identifier?

This makes me want to subdivide this list more formally. One the build gets passing again, i'll merge (maybe need to run it again)

nschonni:
@jbjonesjr the changes to fix the build are in #528
I think splitting it out into subsections like was done for Code for America makes sense.
I'm also not bothered if this one gets closed without merging, just noticed it again when working on the other script stuff

nschonni:
Rebased and fixed merge conflict with `mapnificent`. Travis seems to be stalling on the page check right now

PR #538, opened 2017-01-30
tbs-sct is no longer an organization

jbjonesjr:
Not sure if it was just named something else, or if it is gone completely.

cc/ @nschonni since you added them and they are 🇨🇦 

Let me know if you have a preference or renaming them instead of 🔥 them completely


nschonni:
:shipit: if they go renamed, I've got no idea what the new one is. If I find out, I'll just PR it.

PR #541, opened 2017-02-09
Add "projectestac": Catalonia educational projects

frncesc:


frncesc:
Travis CI fails because there is an error with `bikesafety` (currently in "_Code For America_"), but this has no relation with this PR.

I'm just adding [`projectestac`](https://github.com/projectestac/), an open source organization managed by the ICT in Education Unit of the Department of Education of Catalonia.

PR #543, opened 2017-02-10
Update the fetch-cfa script and resync

nschonni:
Looks like the endpoint was moved, so I updated the url.
After running the script there were a handeful of orgs that we're in there api, so they were moved under Civic Hackers.
The only real deletion with this was the `bikesaftey` org, which appears to be removed.


nschonni:
@jbjonesjr this fixes the build again too

PR #545, opened 2017-02-24
Minor update to the contributing section for Orgs

nschonni:
Highlight that it is for GitHub organizations, and that there are different files now


jbjonesjr:
:+1:

PR #550, opened 2017-03-20
Move cities to their right country

josepjaume:
These Governments should be inside *Catalonia*, not Spain.


jbjonesjr:
thanks @josepjaume

PR #552, opened 2017-03-21
Added LabGCBA

Widcket:
https://github.com/LabGCBA

### DIFF
--- a/_data/governments.yml
+++ b/_data/governments.yml
@@ -4,6 +4,7 @@ Argentina:
   - argob
   - cifasis
   - gcba
+  - LabGCBA
   - inti-cmnb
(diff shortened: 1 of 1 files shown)


jbjonesjr:
Hey @Widcket . @labgcba is a user, not an org, so they won't work. If they want to convert their account to an org, then they can be included. Details can be found here: https://help.github.com/articles/converting-a-user-into-an-organization/

nschonni:
@Widcket it looks like the organizations that the @LabGCBA is apart of might be a good fit https://github.com/gcba-odin and https://github.com/gcba-iris. I might be a better fit under the `civic_hackers.yml` though

PR #556, opened 2017-04-04
Add Sant Cugat and Sabadell's city councils

josepjaume:


jbjonesjr:
TIL: So much Catalonia GitHub stuff.

PR #557, opened 2017-04-20
Update governments.yml

conrad-mac:
Add organsiation account for New Zealand's Social Investment Unit.


jbjonesjr:
cc/ @affrae for awareness

PR #560, opened 2017-04-25
Fixing gov_cloud redirect

jbjonesjr:
redirect_to takes a string, not an array. that also makes total sense.

### DIFF
--- a/docs/gov-cloud.md
+++ b/docs/gov-cloud.md
@@ -1,5 +1,4 @@
 ---
 title: GitHub Enterprise on AWS GovCloud
-redirect_to:
-  - https://enterprise.github.com/aws/gov-cloud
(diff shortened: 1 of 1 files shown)


akinaito:
@jbjonesjr Can you pls let @Kzweng @m0rgane know when you merge (or I guess now they'll get notification 😆 )  Just want to make sure it is a thing before confirming for AWS ad

jbjonesjr:
@akinaito definitely. I hope this is all I need now. we'll see after the next pages build... :crossed_fingers:

jbjonesjr:
WHYYYY.... 

please help @parkr , you're my only hope

nschonni:
looks like it is supposed to be an array https://github.com/jekyll/jekyll-redirect-from#redirect-to

PR #562, opened 2017-04-26
Move custom redirect.html to a new filename to fix compatibility with jekyll-redirect-from

parkr:
https://github.com/jekyll/jekyll-redirect-from/pull/131 added a layout to our redirect page so folks could customize them.

This created an incompatibility with anyone using both `jekyll-redirect-from` and a custom, unrelated layout called `redirect.html`.

This PR should patch up this site.

/cc @jbjonesjr @benbalter


jbjonesjr:
Thanks @parkr . I did not know about the custom redirect model, totally explains how we were using this and overriding the core logic. Thanks for the experience to ferret this out.

Anyone ( @benbalter, @parkr have any issues if I :ship:)

parkr:
:shipit:

benbalter (approved):
:+1: Alternatively, we could kill the custom redirect and just use `jekyll-redirect-from` now that it exists :smile:.

parkr:
> we could kill the custom redirect and just use jekyll-redirect-from now that it exists 😄.

Yeah, though it'd make the process of updating the case study links a little more manual. :shrug:

PR #564, opened 2017-04-27
Update civic_hackers.yml

GovInTrenches:
Adding Chi Hack Night


jbjonesjr:
@GovInTrenches very cool list of projects there!

PR #569, opened 2017-05-11
Update governments.yml

DaveBathnes:
Adding ToonLibraries and LibrariesWest to UK list.  Both local government organisations (doing public library work)

### DIFF
--- a/_data/governments.yml
+++ b/_data/governments.yml
@@ -443,6 +443,7 @@ U.K. Councils:
   - HfdsCouncil
   - kentcc
   - LambethCouncil
+  - LibrariesWest
   - lichfield-district-council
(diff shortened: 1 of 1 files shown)


nschonni:
It looks like the `ToonLibraries` isn't a GitHub organization, but the `LibrariesWest` looks good

jbjonesjr:
Hey @DaveBathnes. @toonlibraries is a user, not an org, so they won't work. If they want to convert their account to an org, then they can be included. Details can be found here: https://help.github.com/articles/converting-a-user-into-an-organization/

DaveBathnes:
Thanks both, silly mistake of mine!  I'll see if they're up for becoming an organisation.

PR #578, opened 2017-07-10
Update government.yml

rez20004:
added Mauritius


_data/governments.yml
```diff
@@ -255,7 +255,9 @@ Luxemburg:
 
 Malaysia:
   - pmo2
-
+Mauritius
```
rez20004: + Mauritius:

PR #581, opened 2017-07-13
chore: remove GSA-Archive which now 404s

nschonni:
Fixes the build


jbjonesjr:
cc/ @gbinal in case you have a different future home

PR #582, opened 2017-07-18
Run fetch-us and fetch-cfa scripts to pull latest updates

nschonni:
Ran the fetch-* and `alphabetize` scripts to cleanup the orgs


nschonni:
Rebased and tweaked the CFA fetch script to retire orgs to the general Civic Hacker list instead of just dropping them when they're no longer in the CFA API

PR #584, opened 2017-07-18
Add Canadian Digital Services org

nschonni:


jbjonesjr:
very happy to see our friends at http://digital.canada.ca/ on board.

Look forward to them adding more organizations all throughout the Candian Govt in the future!

PR #591, opened 2017-08-03
Fix links in footer

sophshep:
The logo was being positioned absolutely, so it was technically overlaid over the links, making it the only clickable thing. 

This fixes it and also makes things more friendly on different screen sizes.

cc/ @benbalter

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -1,22 +1,26 @@
-    <footer class="border-top">
-      <div class="container-lg p-responsive py-4 clearfix">
-        <ul class="list-style-none float-right text-small">
-          <li class="d-none d-lg-inline-block ml-lg-2"><a href="https://help.github.com/articles/github-glossary/" target="_blank">Glossary</a></li>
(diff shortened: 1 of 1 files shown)


_includes/footer.html
```diff
@@ -1,22 +1,26 @@
...
+          <ul class="list-style-none float-right text-small">
+            <li class="d-none d-lg-inline-block ml-lg-2"><a href="https://help.github.com/articles/github-glossary/" target="_blank">Glossary</a></li>
+            <li class="d-none d-lg-inline-block ml-lg-2"><a href="https://help.github.com/articles/github-security/">Security</a></li>
+            <li class="d-inline-block ml-lg-2"><a href="mailto:government@github.com">Contact Us</a></li>
+          </ul>
 
-        <a href="/" class="text-center position-absolute left-0 right-0">
-          {% octicon mark-github height:24 class:"fill-gray-light d-inline" aria-label:github-logo %}
-        </a>
+          <a href="/" class="text-center">
```
nschonni, line 10: I noticed that the education.github.com doesn't use a link at all, not sure if that makes sense

benbalter (approved):
Thanks for cleaning this up!

PR #593, opened 2017-08-10
Update governments.yml

liammcmurray:
Remove Environment Agency as all github repos are under Defra organisation now.


nschonni:
Are they going to remove the org too, or just leave it to prevent squatters :)

liammcmurray:
@nschonni I'll ask!
(time passes)

Hey - yeah someone here created it to stop someone else creating it. Not sure why, but it's not going to ever be active - all our code is under Defra.

⚡️

liammcmurray:
Despite this being kept as mentioned above, it should be removed from an official list as it will never be active and is misrepresentative.

PR #595, opened 2017-08-18
Remove non-governmental organization from Chile

aarellano:
This project has no links with any government agency


nschonni:
I'd say move it to the civic_hackers, but it really doesn't have any repos either 😛

PR #599, opened 2017-08-21
Update governments.yml

rasky:
Add the organization owned by Team Digitale (part of the Italian Government).


jbjonesjr:
thanks @rasky . You're all set here.

PR #606, opened 2017-09-13
Add sipf from french polynesia government

sitle:


sitle:
@benbalter : please let me know if you need more informations. Thx.

_data/governments.yml
```diff
@@ -317,6 +317,9 @@ Poland:
   - coi-gov-pl
   - Ministerstwo-Cyfryzacji
 
+Polynésie Française:
```
nschonni: I think this should be `French Polynesia` since the other countries keep the English titles.
Not sure if this wouldn't go under France though https://en.wikipedia.org/wiki/French_Polynesia (sorry for wiki as a source)

sitle: I change it to english to be compliant (even if we are frenchies :p). Thx for your review.

PR #607, opened 2017-10-03
Change NCSC org to better one

annashipman:
NCSC have two organisations on GitHub; the one linked to here has no public repos. The one I've changed it to in this commit has many public repos, so this is a better one to link to.

### DIFF
--- a/_data/governments.yml
+++ b/_data/governments.yml
@@ -425,7 +425,6 @@ U.K. Central:
   - ministryofjustice
   - naouk
   - nationalarchives
-  - ncscuk
   - OfqualGovUK
   - ONSdigital
   - openglasgow
@@ -437,6 +436,7 @@ U.K. Central:
   - UKGovernmentBEIS
   - UKHomeOffice
   - UKLocation
+  - ukncsc
 
 U.K. Councils:
   - BarnsleyCouncil


jbjonesjr:
Hey @annashipman , quick question then: Why does the NCSC still have two orgs (`ncscuk` & `ukncsc`)? Is it purely based on some GitHub billing mechanism (aka, the one with public repos is a free plan and open to everyone, the other one has private repos so it's membership is much tighter?

Clearly multiple orgs can be an anti-pattern, and I'd love to understand if it's a GitHub-influenced anti-pattern in this case.

### Note:
Merging anyway, as that doesn't effect the current state and what is best to showcase

annashipman:
@jbjonesjr I have no idea, I'm afraid, I don't work for NCSC, I work elsewhere in UK government (GDS) as the open source lead.

annashipman:
Thanks for the merge, it will make it easier for people to see what code is public! 👍

jbjonesjr:
@annashipman totally understand, different agencies and priorities :) Wasn't sure if you knew the situation over there (or if GDS had a policy that NCSC was perhaps following for this stuff). I'll keep my eyes out for an NCSC contact and maybe get to the bottom of this "one day". Just want to make sure everyone is getting what they need!

PR #613, opened 2017-10-17
chore: Los Angeles org moved?

nschonni:
It looks like it has been deleted

### DIFF
--- a/_data/governments.yml
+++ b/_data/governments.yml
@@ -502,6 +502,7 @@ U.S. City:
   - cno-opa
   - CORaleigh
   - CUMTD
+  - CityOfLosAngeles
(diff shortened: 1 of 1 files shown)


jbjonesjr:
probably cc/ @hunterowens who i don't think is in this org (so he won't get the ping, but if anyone knows about those LA things...)

hunterowens:
Hey @jbjonesjr! Sorry for the late notice. We've merged the two LA associated orgs (datala/CityOfLosAngeles) onto a single organization name (CityOfLosAngeles)

jbjonesjr:
@hunterowens thanks for the update. appreciate it

PR #612, opened 2017-10-17
Update governments.yml

Webspook:

PR #614, opened 2017-10-18
Update australian government orgs

maxious:


jbjonesjr:
so many Australian updates today! (#612). Thanks for sharing this @maxious .  Also interesting the multiple Antartica orgs!

The issue with this PR is that `CAPITAmodel` is currently a user, and not an org. As per the Government Best Practices, we suggest they be an [Organization, not a user's account](https://government.github.io/best-practices/account-management/#organization-or-user). If you want to convert your account to an org, details can be found here: https://help.github.com/articles/converting-a-user-into-an-organization/

cc/ @affrae @samuelhunt

maxious:
Aww, it is a very cool repo. 
In order to not block recognition for our way down under contributors, I've removed CAPITAmodel from this PR.

jbjonesjr:
@maxious thanks for the update. I agree, CAPITAmodel is actually a very useful tool provided out to the public (who doesn't want to do some tax modeling??). If they get it into an org, i'd be happy to have it added back in.

PR #615, opened 2017-10-18
Adding AMA Portugal to governments.yml

miguelantoniosantos:
@amagovpt - https://www.ama.gov.pt/


jbjonesjr:
obrigado @Cotemero 🙇

PR #631, opened 2017-11-16
Request The bantenprov in government.github.com list

andri-sudarmawijaya:


nschonni:
@andri-sudarmawijaya and @bantenprov you seem to both be trying to add this GitHub user. You'll need to convert the GitHub user to an Organization before the build will work and it can be added. Details can be found here: https://help.github.com/articles/converting-a-user-into-an-organization/

PR #633, opened 2017-11-20
Update governments.yml

KyleStr:
Adding the Inter-American Development Bank's account, where we are consolidating code and digital tools that governments and others can use to address their development challenges, under the Code for Development project.

### DIFF
--- a/_data/governments.yml
+++ b/_data/governments.yml
@@ -226,6 +226,7 @@ India:
 International:
   - IATI
   - ogpl
+  - EL-BID
 
 Isle of Man:
   - Isle-Of-Man-Government


jbjonesjr:
Hey @KyleStr , is the IADB federally/govt funded? I think that it might fit more in the `civic-hackers` category perhaps?

KyleStr:
Hi @jbjonesjr, the IDB is completely owned and funded by its 48 member governments.    We are an international organization, a multi-government public sector development agency, and are the leading source of development financing for Latin America and the Caribbean.   We are not a civic-hackers group as we are civil servants.  Does that respond to your doubts?   You can read more at www.iadb.org<http://www.iadb.org>.  Please let me know if you have additional questions about the request.

jbjonesjr:
cool. you convinced me! merging ahead

PR #626, opened 2017-11-08
spelling: based

jsoref:
Generated by https://github.com/jsoref/spelling `f`; to maintain your repo, please consider `fchurn`

PR #639, opened 2017-12-20
Typo, "bassed" should be "based".

dangowans:
text-bassed should be text-based
flash-bassed should be flash-based


dangowans:
Looks like this was already fixed in #626 .  Oops.

jbjonesjr:
no opps, thanks for surfacing it (again) @dangowans

Issue #641, opened 2017-12-26
Overflow issue


max107:
![screen shot 2017-12-26 at 13 36 07](https://user-images.githubusercontent.com/862272/34354796-db9e6fe2-ea41-11e7-88cb-45f91130d3de.png)
![screen shot 2017-12-26 at 13 36 17](https://user-images.githubusercontent.com/862272/34354799-df629aae-ea41-11e7-8b22-ef6fba4b5d6a.png)

Solution:

remove negative margins:

```scss
.featured-orgs {
  max-height: 348px;
  overflow: hidden;
  text-align: center;
}
```

jbjonesjr:
any thoughts on this (and the related #642) @sophshep ? Wasn't sure if there was a reason for the negative margins.

### Before:
![image](https://user-images.githubusercontent.com/533967/34463883-40133ee0-ee38-11e7-8d61-d50311f00b5d.png)
Logos are overflow/cropped.

### After: 
![image](https://user-images.githubusercontent.com/533967/34463884-59727586-ee38-11e7-917f-57a346260021.png)

Nicely see them all.

I kind of like the `cropped` look, but some are cropped too much (see the 18F logo for example) where you can't actually tell what it is.

jbjonesjr:
fixed by #679

Issue #637, opened 2017-12-18
Organisation logos are shown twice in the "Join the Revolution" section


ilyapopov:
![screenshot-2017-12-18 github and government](https://user-images.githubusercontent.com/1054514/34116526-2d74901e-e419-11e7-9723-c96d980c3383.png)

alex-page:
The scrolling on the page is also broken due to the second group of logos not being wrapped in the `overflow:hidden;` div.

jbjonesjr:
yep. fixed in #679

PR #647, opened 2018-01-04
Fix for #637 and #641 issues

Violettica:
This should both remove the negative margins (#641) and the duplicated nonsense on #637.

### DIFF
--- a/assets/css/custom.scss
+++ b/assets/css/custom.scss
@@ -33,8 +33,8 @@ body {
 }
 
 .featured-orgs {
-  margin-right: -55px;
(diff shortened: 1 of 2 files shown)


Violettica:
I apologize for the excessive roundabout commits, it took a lot of prodding to find the actual source of the problem.

miaguila0710:
My soon have a big problems and need this very urgent

jbjonesjr:
fixed by #679 .

Thanks for your investigation @Violet-Maverick , this was helpful in focusing us to think about the intended solution internally.

PR #650, opened 2018-01-17
chore: Remove codeforraleigh

nschonni:
Organization appears to be gone and is breaking the build


jbjonesjr:
http://ncopenpass.com/

> NC Open Pass is a partnership between Code for Cary, Code for Durham, Code for Raleigh, and NC DataPalooza—making our innovative events in the Triangle available together on one ticket.

PR #672, opened 2018-03-26
Exclude node_modules directory from Jekyll build

MrChrisW:
The primer modules in use all have README.md files which are publicly accessible and have been indexed.

+ https://government.github.com/node_modules/primer-buttons/

image

This PR removes them from the Jekyll build.


nschonni:
It seems like it's used below on L26 for the `load_paths` not sure if that would cause an issue

_config.yml
```diff
@@ -7,7 +7,7 @@ relative_permalinks: false
 analytics_account: "UA-3769691-28"
 
 # build settings
-exclude: ["script", "vendor", "bower.json", "Gemfile", "Gemfile.lock", "Rakefile", "readme.md", "package.json"]
+exclude: ["script", "vendor", "bower.json", "Gemfile", "Gemfile.lock", "Rakefile", "readme.md", "package.json", "node_modules"]
```
jbjonesjr, line 10: I'm super 👌 with this, but want to confirm that adding this into the excludes doesn't screw up the `load_paths` below (not sure how it could be loaded if excluded, but then again, I don't Jekyll, so.....)

jbjonesjr (approved):

_config.yml
jbjonesjr: Hey @MrChrisW , trusting you here.

jbjonesjr:
cc/ @sophshep would still love a review from someone on the `site` team to make sure we're not doing anything weird.

sophshep (approved):
Tested on staging: https://github-5bc252ad4f.drafts.github.io/government.github.com/

Everything works as expected.

jbjonesjr:
thx @sophshep

_config.yml
MrChrisW: @jbjonesjr Thanks 💜

PR #679, opened 2018-04-03
Fix repeating logos & scroll bug

sophshep:
@kimestoesta noticed that the homepage had a horizontal scroll bug: 

image

While debugging this, I noticed that the community logos matrix was shown twice:

image

This PR fixes both.

### DIFF
--- a/index.html
+++ b/index.html
@@ -136,12 +136,7 @@ permalink: /
       {% endfor %}
     </div>
   </div>
-
-<div class="full-screen featured-orgs">
-  {% for org in site.featured_orgs %}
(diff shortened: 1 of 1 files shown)


jbjonesjr:
@sophshep it looks like this PR so far only fixes the duplicate logos section.  I assume you meant to commit a index.scss file as well?

sophshep:
> I assume you meant to commit a index.scss file as well? 

Removing this block should fix both. Are you seeing it still @jbjonesjr?

jbjonesjr:
i looked, but didn't see the change. i just did a clean clone, and all looks well now. ¯\_(ツ)_/¯

jbjonesjr (approved):

PR #724, opened 2018-10-10
Add Kanton Bern (Switzerland)

msiegenthaler:
We, the canton of Berne (Switzerland) have set up an official open source "platform" on github and would like to add it to the list.

I (@msiegenthaler) am a external consultant, if you need to confirm with someone having an official account (`@*.be.ch`) then please contact @Thomas-Joos, who is also the owner of the organization.

Thanks for the contribution! **Below** is a template that can make it easier when submitting your Organization:

**Your Organization**: Kanton Bern
**GitHub Organization url**: @kanton-bern
**Country/Locality**: Bern, Switzerland

Inclusion requirements:
- [X] Referenced account is an [Organization](https://github.com/github/government.github.com#add-organization) not a User^
- [X] If this is a government project, your Organization's description should include a reference to your country/agency.
- [X] Make sure your Organization includes at least one public repository
- [X] Please also include a URL for your organization that links back to your organization or agency homepage.
 
^ If you want to convert your account to an org, details can be found here: https://help.github.com/articles/converting-a-user-into-an-organization/
  
 :heart: GitHub Government Contributors

PR #723, opened 2018-10-04
Add new organisations under the @ANSSI-FR umbrella

yaperez-anssi:
CLIP OS has been published at https://clip-os.org
Wookey has been published at https://wookey-project.github.io/

**Your Organization**: _ANSSI-FR_ 
**GitHub Organization url**: _@clipos_ _@clipos-archive_   _@wookey-project_
**Country/Locality**: _Paris, France_

PR #720, opened 2018-10-02
Update governments.yml

HealthPDU:
Include 'Department for Work and Pensions' to UK Central

**Your Organization**: Department for Work and Pensions
**GitHub Organization url**: https://github.com/dwp
**Country/Locality**: U.K. Central

PR #717, opened 2018-09-21
Agregar al Gobierno de Tamaulipas

jpgamez:
Thanks for the contribution! **Below** is a template that can make it easier when submitting your Organization:

**Your Organization**: _Gobierno del Estado de Tamaulipas_  
**GitHub Organization url**: _@tamaulipas_  
**Country/Locality**: _Tamaulipas, México_

Inclusion requirements:
- [x] Referenced account is an [Organization](https://github.com/github/government.github.com#add-organization) not a User^
- [x] If this is a government project, your Organization's description should include a reference to your country/agency.
- [x] Make sure your Organization includes at least one public repository
- [x] Please also include a URL for your organization that links back to your organization or agency homepage.
 
^ If you want to convert your account to an org, details can be found here: https://help.github.com/articles/converting-a-user-into-an-organization/
  
 :heart: GitHub Government Contributors

PR #725, opened 2018-10-14
Fix errors causing build to fail the rubocop checks.

thomasgohard:
This is causing all PRs to fail since #715. Currently affects PRs #715, #717, #718, #720, #721, #722, #723, and #724.

The root cause is likely https://github.com/rubocop-hq/rubocop/pull/6235, which turned on a check that used to be disabled by default.


jbjonesjr:
this is true. thanks @thomasgohard

PR #748, opened 2019-01-24
improve our stats

jbjonesjr:
The front page doesn't have correct GitHub.com information. This attempts to update that.


jbjonesjr:
I'll also go through and update some other messaging while i'm at it. cc/ @elstudio if you have ideas.

PR #778, opened 2019-10-03
rm 404 link

mlinksva:


mlinksva:
I'm baffled why bundler running in CI is pulling very old versions of pages and jekyll, which I think is what is causing the build failure.

PR #780, opened 2019-10-04
Fix build

nschonni:
Think it might be something odd with an old gem release, but pinning the pages gem to a range larger than the one it was installing (103) seems to work


mlinksva:
Thanks! I also found odd in https://github.com/github/government.github.com/pull/778#issuecomment-537788886 ... let's try your constraint hack.

PR #782, opened 2019-10-05
The _data is unrestricted, CC0-1.0

### LINKED ISSUE #755: Add licence to the _data directory
Add a suitable licence to the _data directory to allow it to be reused for research purposes

mlinksva:
Fixes #755


nschonni:
Think an actual LICENSE.txt or .md would be better

mlinksva:
@nschonni normally I'd agree but this is just to clarify the status of files in one directory. Dropping a LICENSE file in `_data` would still merit some explanation.

nschonni:
I was thinking the license would be in the root rather than under the _data

mlinksva:
Exactly: in root would be confusing, because (at this point) it is only intended to apply to `_data`.

PR #818, opened 2020-08-05
Remove "usinterior" org

evantayloryates:
There is no public GitHub org with the name "usinterior"


nschonni:
Think this got fixed in https://github.com/github/government.github.com/pull/826 for the new name, after a separate one I did dropped it

mlinksva:
Already removed, thanks for noticing and the PR though!

PR #830, opened 2020-09-05
chore: Add nightly sync for CFA orgs

nschonni:
This will probably need the resolution of some missing brigades that had to be manually added back in the previous PR

### DIFF
--- a/.github/workflows/sync-cfa.yml
+++ b/.github/workflows/sync-cfa.yml
@@ -1,0 +1,35 @@
+name: Run CFA fetch
+
+on:
+  schedule:
+    - cron: "0 0 * * *" # Run nightly at midnight
+
+jobs:
+  build:
+
+    runs-on: ubuntu-latest
+    if: github.repository_owner == "github"
+
+    steps:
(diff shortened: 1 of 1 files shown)


PacMani3 (approved):
~~Tosha~~

.github/workflows/sync-cfa.yml
```diff
@@ -0,0 +1,35 @@
...
+        ruby-version: 2.6.x
+    - name: Setup Dependencies
+      run: |
+        gem install bundler
+        bundle install --jobs 4 --retry 3
+    - name: Fetch latest CFA changes
+      run: bundle exec script/fetch-cfa
+    - name: Ensure that all are still orgs
+      run: bundle exec script/ensure-orgs
+    - name: Alphabetize changes
```
mlinksva, line 27: If any of the above fail, say network failures, does the below automatically push a commit deleting all civic hacking orgs? Or does processing stop?

nschonni: Process should stop, because "continue on error" isn't set on any of them

PR #851, opened 2020-11-19
feat: Update workflows to use gem caching and latest Ruby 2.7.

### LINKED ISSUE #750: TODO: Configure caching for fast building on CI
Description and details here: https://github.com/gjtorikian/html-proofer#configuring-caching and https://github.com/gjtorikian/html-proofer#caching-with-travis

elstudio:
Update workflows to use gem caching, by using the more full-featured ruby/setup-ruby action.

This action combines the functionality of both actions/setup-ruby and actions/cache — and makes it a bit easier to specify ruby versions, too.

Now specifying the ruby version in our .ruby-version file: Updated to 2.7.2.

Fixes #750.


.github/workflows/build.yml
```diff
@@ -10,11 +10,9 @@ jobs:
     steps:
     - uses: actions/checkout@v2
     - name: Set up Ruby 2.6
-      uses: actions/setup-ruby@v1
+      uses: ruby/setup-ruby@v1
       with:
-        ruby-version: 2.6.x
+        ruby-version: 2.6
```
nschonni: Looks like these can be dropped now and it will read the `.ruby-version` https://github.com/ruby/setup-ruby#supported-version-syntax

nschonni, line 15: Looks like these can be dropped now and it will read the `.ruby-version` https://github.com/ruby/setup-ruby#supported-version-syntax

.github/workflows/rubucop.yml
```diff
@@ -22,11 +22,9 @@ jobs:
...
       with:
-        ruby-version: 2.6.x
+        ruby-version: 2.6
+        bundler-cache: true
     - name: Rubocop
-      run: |
-        gem install bundler
-        bundle install --jobs 4 --retry 3
-        bundle exec rubocop -D -S
+      run: bundle exec rubocop -D -S -a
```
nschonni: I don't think autocorrect makes sense for the CI. The CI will pass, even if there are issues in the checked in code as long as they are fixable

.github/workflows/build.yml
elstudio: True. However, in .ruby-version we must specify the full 2.6.6 version — since rbenv requires a full version (it can’t work with 2.6). 

I think it’s better to leave the major version in our workflows, which will automatically use the latest 2.6 patch.

Folks building locally with rbenv will still use the .ruby-version — and so I’ve matched that for now.

Downside of that approach is that it repeats ourselves, kinda. But I think it’s better to automatically get patch releases in our workflows. What do you think?

.github/workflows/rubucop.yml
elstudio: Oh yeah! Good catch!

elstudio: Good side about my making dumb mistakes like this — gem caching really works!

.github/workflows/build.yml
nschonni: Fair enough. I think 2.7.1 is the latest for the pages setup though. I usually peak over at https://github.com/github/pages-gem/ when I'm looking to bump that

.github/workflows/rubucop.yml
nschonni: I hadn't seen that option before, so now I'm going to go update a few workflows for that cache seeing 😄

.github/workflows/build.yml
elstudio: Yeah. And 2.7 will run a little faster. Tests do pass running locally with 2.7, so that looks okay.

Thinking about this a little more, since we have multiple workflows here, why don’t we leave the ruby version specification to .ruby-version. That way we have just one file to keep up to date, rather than 4 workflows, too.

So I’ll push that commit in a second here. You okay with that approach?

nschonni: Yup, that works for me. I think I came across somewhere that had an automated way to pull down the info from an API to keep it in sync, but I don't recall were now

nschonni: Looks like you cleaned up the other files, but not this one 😉

elstudio: Whoops. Fixed now

PR #894, opened 2021-12-22
Try to fix #881

### LINKED ISSUE #881: Broken Link on government.github.com
Your webpage https://government.github.com has a couple links to the URL below.
https://government-community.githubapp.com/

This link produces an error.

girolamodaschio:
Manually redirecting (https://government-community.githubapp.com/) to (https://github.com/government/welcome)


nschonni:
The app is still required (although i guess broken) to get access to the private repos in the org

mlinksva:
I think it'd be better to remove the links and their link text entirely until they are fixed rather than linking to instructions which also won't work. I'd welcome a PR doing this which would be easy to revert if/when the links are fixed.

PR #895, opened 2021-12-30
Fixing #881

girolamodaschio:
As suggested here: https://github.com/github/government.github.com/pull/894

### DIFF
--- a/index.html
+++ b/index.html
@@ -167,11 +167,7 @@ permalink: /
         <h3 class="alt-h3 mb-2">Sign up for an account</h3>
         <p class="text-gray"><a href="https://github.com/signup/free" target="_blank">Sign up</a> for an individual GitHub account for free.</p>
       </div>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -177,7 +176,6 @@ <h3 class="alt-h3 mb-2">Create an organization</h3>
     <div class="mb-md-6 clearfix gutter-spacious">
       <div class="mb-4 col-md-6 float-left">
         <h3 class="alt-h3 mb-2">Join the Peer Group</h3>
-        <p class="text-gray">Join the <a href="https://government-community.githubapp.com/">semi-private group</a> of government employees using GitHub to share best practices for software collaboration.</p>
```
mlinksva, line 180: I think the entire enclosing div abut the peer group needs to go.

mlinksva (changes requested):

PR #897, opened 2022-01-04
Remove Heroku and Eloqua Web Tracking

mike421453:
Removing [Heroku reference](https://github.com/orgs/github/teams/engineering/discussions/607) (and a few other things since our hands are dirty)


mike421453:
@mlinksva - Hi, would you happen to know the status of https://government-community.githubapp.com/ ? (or do you know who does?)

There are a number of references to it in this site and since it appears to be down/missing, it is causing [a failure of the checks](https://github.com/github/government.github.com/runs/4707399731?check_suite_focus=true) required to deploy this PR.  For background, I am removing some old javascript that are no longer functional and addressing a Bug Bounty.

cc: @jeffreyguerra

PR #921, opened 2022-04-25
Add ~~berlinonline and~~ digitalservice4germany

codedust:


mlinksva:
Is berlinonline a government organization? At a glance it looks like a private consultancy that has government clients, which would probably be out of scope for this catalog. Happy to be wrong!

codedust:
You are absolutely right. Sorry. I'll update the PR.

https://www-berlin-de.translate.goog/rbmskzl/aktuelles/pressemitteilungen/2021/pressemitteilung.1102027.php?_x_tr_sl=auto&_x_tr_tl=en&_x_tr_hl=de&_x_tr_pto=wapp

PR #970, opened 2023-01-07
try using GITHUB_TOKEN in CI and obtaining fewer 429s etc?

mlinksva:


nschonni:
Probably need to do this in the `scripts` too

mlinksva:
Yeah, will add elsewhere if/when this works.

Rakefile
```diff
@@ -4,9 +4,14 @@ require 'html-proofer'
...
-                              hydra: { max_concurrency: 10 },
-                              url_ignore: [%r{https://developer.github.com}, %r{https://docs.github.com}, %r{https://help.github.com}],
-                              http_status_ignore: [429]).run
+  proofer = HTMLProofer.check_directory(
+    './_site/',
+    check_html: true,
+    hydra: { max_concurrency: 10 },
+    url_ignore: [%r{https://developer.github.com}, %r{https://docs.github.com}, %r{https://help.github.com}],
+    http_status_ignore: [429])
+  proofer.before_request do |request|
```
nschonni: might want to check if the value is set, so the local script works if they haven't set this up. Maybe the empty header isn't an issue though

Rakefile
```diff
@@ -4,9 +4,14 @@ require 'html-proofer'
...
+    request.options[:headers]['Authorization'] = "Bearer #{GITHUB_TOKEN}" if request.base_url == "https://github.com"
```
nschonni: I'm bad at ruby, but does it need to be something like `ENV["GITHUB_TOKEN"]`

mlinksva: Me too, just noticed due to failing CI :)

mlinksva:
Mysteries:
* obviously not avoiding 429
* ~~htmlproofer config should ingore them~~ (updated htmlproofer config for new version)
* ~~Not sure why 2 workflows aren't running~~ (they only run if changes in `_data`, by design!)

nschonni:
Other approach might be to try getting it working on one of the other scripts with the native Typeous config, then use that for HTMLProofer https://github.com/gjtorikian/html-proofer/wiki/How-to-configure-Typhoeus

mlinksva:
Not clear to me passing token is working, but other fixes here useful. Tests pass for 1st time in awhile.

PR #1069, opened 2023-02-04
Remove vest-agder

psjnek:
vest-agder was seemingly deleted and the link now results in a 404.


mlinksva:
Thanks for this @WebSnke

PR #1126, opened 2023-05-02
Remove deleted organiszations

psjnek:


_data/governments.yml
```diff
@@ -484,9 +484,6 @@ Singapore:
   - moexmen
   - opengovsg
 
-Slovakia:
-  - slovak-egov
```
nschonni: https://github.com/slovak-egov still exists. Maybe you're hitting an API limit rather than actual removed orgs

mlinksva:
Appears Code for Hawaii moved as a result of https://statescoop.com/code-for-america-local-brigades-ending/ so re-added new org.

Issue #1168, opened 2023-06-09
Set up locally instructions failing


ras44:
Thank you for building this repo!

I'm working through the **Set up locally** instructions at: 

https://github.com/github/government.github.com#to-set-up-locally

On ubuntu, which matches the `.github/workflows/build.yml` I can build and serve with the following:

```
sudo apt update
# following https://jekyllrb.com/docs/installation/ubuntu/
sudo apt-get install ruby-full build-essential zlib1g-dev
echo '# Install Ruby Gems to ~/gems' >> ~/.bashrc
echo 'export GEM_HOME="$HOME/gems"' >> ~/.bashrc
echo 'export PATH="$HOME/gems/bin:$PATH"' >> ~/.bashrc
source ~/.bashrc
gem install jekyll bundler
# install node.js
sudo apt install nodejs
# following To Set up Locally instructions
git clone https://github.com/github/government.github.com.git
cd government.github.com
script/bootstrap
script/server
```

If I've understood things correctly, the missing requirement from the instructions is the installation of a JavaScript runtime.

Not sure if there might also be an option to use the docker image that is pulled by the GitHub actions process?

```
/usr/bin/docker pull ghcr.io/actions/jekyll-build-pages:v1.0.7
```

Thanks so much for your work and guidance 🙏

mlinksva:
This was addressed by #1169

PR #1169, opened 2023-06-09
Mention JavaScript runtime requirement #1168

ras44:
Addresses #1168

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -40,7 +40,7 @@ You can take all the files of this site and run them just on your computer as if
 * [Jekyll](https://jekyllrb.com/)
 * [Ruby](https://www.ruby-lang.org/en/)
 * [Git](https://git-scm.com/)
-* A JavaScript runtime (ie. [node.js](https://nodejs.org/en))
+* A JavaScript runtime ([node.js](https://nodejs.org/en))
 
 _If you have installed [GitHub Desktop](https://desktop.github.com), Git was also installed automatically._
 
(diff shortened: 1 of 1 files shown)


readme.md
```diff
@@ -40,6 +40,7 @@ You can take all the files of this site and run them just on your computer as if
 * [Jekyll](https://jekyllrb.com/)
 * [Ruby](https://www.ruby-lang.org/en/)
 * [Git](https://git-scm.com/)
+* [A JavaScript runtime (ie. node.js)](https://github.com/rails/execjs)
```
mlinksva: I think it's confusing to link to execjs which is not used.

mlinksva (changes requested):

readme.md
ras44: I agree.  I got that link from the error message that Ruby/Jekyll outputs when it can't find a JavaScript runtime.  I think they reference that in order not to favor one runtime over another.  Not sure if there's a preferred option?  I installed node.js and had no issues.

ras44: @mlinksva FYI here's an example of execjs being referenced in the error output from jekyll:

- https://github.com/jekyll/jekyll/issues/2327#issue-32978237

(see first line of stacktrace)

The issue goes on to describe a dependency on coffeescript which is going to be deprecated in jekyll V3, and it looks like in [some cases](https://github.com/jekyll/jekyll/issues/2327#issuecomment-54740996) nodejs was the only solution that worked.

The error message is also noted here:

jekyll docs: 

- https://github.com/jekyll/jekyll/blob/master/docs/_docs/troubleshooting.md#could-not-find-a-javascript-runtime-execjsruntimeunavailable

ras44:
Just changed the link to nodejs.  I understand from the issue descriptions that some people are concerned about the weight of that dependency, but perhaps in the context of avoiding execjs confusion in this README, this could be a simple solution.

PR #1214, opened 2023-08-01
Remove deleted users

psjnek:


_data/civic_hackers.yml
```diff
@@ -227,7 +226,6 @@ Code For America:
   - codeforfoco
   - codeforftl
   - codeforgiessen
-  - codeforgreenville
```
mlinksva, line 230: Looks like this has been replaced with hackgvl

PR #1215, opened 2023-08-01
Remove orgs without profile picture

psjnek:
I just removed the orgs that do not have a profile picture and replaced them with orgs that do have one.

This is what the list currently looks like:
image


mlinksva:
Thanks, good idea!

PR #1239, opened 2023-09-28
Remove OGPL

psjnek:
OGPL does not really use GitHub that much anymore and it only includes two repositories.


mlinksva:
I'd rather not make that kind of call. If they delete the org we can remove it.

I can see eventually we might want to collect some info on activity and have a different criteria for inclusion or highlighting in some way.

PR #1287, opened 2024-10-19
Remove Canada:MTS-STM

benhovinga:
@MTS-STM is an inactive/empty org with broken website link. Entry doesn't provide any value to this list.

### DIFF
--- a/_data/governments.yml
+++ b/_data/governments.yml
@@ -134,6 +134,7 @@ Canada:
   - cse-cst
   - CSPS-EFPC-DAAN
   - CybercentreCanada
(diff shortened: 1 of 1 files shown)


nschonni:
Looks like maybe it moved over to https://github.com/DTS-STN, but that could be a separate PR

benhovinga:
It's hard to tell because the names are so different and there is no move/forwarding information on the Org's GitHub page.

@MTS-STM should still be removed in any case due to being an inactive Org with no public repos.

@DTS-STN is not in the list and should be a separate PR using the template to get added.

